PR #8, opened 2015-12-16
Added Arista support

LuvCurves:


lib/beaker-hostgenerator/data/vmpooler.rb
```diff
@@ -3,6 +3,10 @@ module Data
     module Vmpooler
 
       OSINFO = {
+        'arista-32' => {
```
joshcooper: I think we want `arista4-32`? Presumably we might have `arista5`, etc at some point?

PR #9, opened 2015-12-16
(QENG-3275) Add support for 32-bit puppet on 64-bit windows

joshcooper:
This PR ensures `ruby_arch` defaults to the correct value based on the Windows OS architecture as the setting is required to run puppet, facter and pxp-agent acceptance on Windows hosts, and allows the caller to request 32-bit Puppet on 64-bit Windows, e.g. `windows2012r2-6432a`. Windows is unique in that there are three combinations, and we currently support all three:

* 32-bit Puppet on 32-bit Windows
* 32-bit Puppet on 64-bit Windows
* 64-bit Puppet on 64-bit Windows

### DIFF
--- a/lib/beaker-hostgenerator/data/vmpooler.rb
+++ b/lib/beaker-hostgenerator/data/vmpooler.rb
@@ -284,6 +284,11 @@ module BeakerHostGenerator
           'template' => 'win-2003-x86_64',
           'ruby_arch' => 'x64'
         },
+        'windows2003-6432' => {
+          'platform' => 'windows-2003-64',
+          'template' => 'win-2003-x86_64',
+          'ruby_arch' => 'x86'
+        },
         'windows2003r2-32' => {
           'platform' => 'windows-2003r2-32',
           'template' => 'win-2003r2-i386',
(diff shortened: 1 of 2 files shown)


lib/beaker-hostgenerator/data/vmpooler.rb
```diff
@@ -281,59 +281,103 @@ module Vmpooler
...
+        'windows2012r2-6432' => {
+          'platform' => 'windows-2012r2-64',
+          'template' => 'win-2012r2-x86_64',
+          'ruby_arch' => 'x86'
         },
         'windows7-64' => {
           'platform' => 'windows-7-64',
-          'template' => 'win-7-x86_64'
+          'template' => 'win-7-x86_64',
+          'ruby_arch' => 'x64'
```
joshcooper: We do, but we haven't historically tested that combination AFAIK. And we're moving towards deprecating and removing support for 32-bit Puppet on any 64-bit system, so I'd be inclined to not add it, unless there's a specific need to. /cc @Iristyle

justinstoller: Cool, I mean the data is syntactically correct, as long as it's the correct data. I've been pinged on what looks like a re-evaluation of our Windows 32 bit testing matrices elsewhere, so I wanted to broach the subject.

I'm also fine, if getting a definitive answer is going to take awhile, to merge this and update it later if needed.

joshcooper:
I need to refork the repo now that's it's public. I'll close this and open a new PR

PR #10, opened 2015-12-18
(QENG-3275) Add support for 32-bit puppet on 64-bit windows

joshcooper:
This PR ensures `ruby_arch` defaults to the correct value based on the Windows OS architecture as the setting is required to run puppet, facter and pxp-agent acceptance on Windows hosts, and allows the caller to request 32-bit Puppet on 64-bit Windows, e.g. `windows2012r2-6432a`. Windows is unique in that there are three combinations, and we currently support all three:

* 32-bit Puppet on 32-bit Windows
* 32-bit Puppet on 64-bit Windows
* 64-bit Puppet on 64-bit Windows

### DIFF
--- a/lib/beaker-hostgenerator/cli.rb
+++ b/lib/beaker-hostgenerator/cli.rb
@@ -101,6 +101,12 @@ Usage: beaker-hostgenerator [options] <layout>
 
       puts "\n"
 
+      puts "valid beaker-hostgenerator architectures:"
+      puts "   32   => 32-bit OS"
+      puts "   64   => 64-bit OS"
+      puts "   6432 => 64-bit OS with 32-bit Puppet (Windows Only)"
+      puts "\n"
+
       roles = BeakerHostGenerator::Utils.get_roles
       puts "valid beaker-hostgenerator host roles:  "
       roles.each do |k,v|
--- a/lib/beaker-hostgenerator/data/vmpooler.rb
+++ b/lib/beaker-hostgenerator/data/vmpooler.rb
@@ -284,6 +284,11 @@ module BeakerHostGenerator
           'template' => 'win-2003-x86_64',
           'ruby_arch' => 'x64'
         },
+        'windows2003-6432' => {
+          'platform' => 'windows-2003-64',
+          'template' => 'win-2003-x86_64',
+          'ruby_arch' => 'x86'
+        },
         'windows2003r2-32' => {
           'platform' => 'windows-2003r2-32',
           'template' => 'win-2003r2-i386',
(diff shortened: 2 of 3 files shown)


joshcooper:
Supercedes #9

joshcooper:
@justinstoller comment from #9:

> Cool, I mean the data is syntactically correct, as long as it's the correct data. I've been pinged on what looks like a re-evaluation of our Windows 32 bit testing matrices elsewhere, so I wanted to broach the subject.

I'm not sure I totally understand your question.. So we've always installed 64-bit Puppet on 64-bit Windows and 32-bit Puppet on 32-bit Windows. The addition of `ruby_arch` just reaffirms what beaker already does, and is required by acceptance tests that need to know which puppet arch is installed.

I didn't add the 32-bit Puppet on 64-bit Windows combinations for desktop platforms, because FOSS doesn't test them. I'm not sure PE does, but would be easy to add if they do.

justinstoller:
That wasn't so much a question as an explanation as to why I asked the previous question:
> Do we support 32bit Puppet on 64[bit] Desktop systems?

I had just seen discussions around: https://docs.google.com/document/d/1Ot6zW5aEX2P8s8JVppokXLOJ851c3do8XzwdCVHoSkQ/edit#heading=h.6yg2l174jlfe

That made me wonder if we were going to be reworking our test matrices. And if in reworking them possibly including 32 bit puppet on 64 bit desktop OSes. And if that was a possibility, then would we want to include the same 32bit on 64bit hostgenerator option strings for desktop OSes that we're including for server OSes in the PR?

I think if there's confusion around that though I'm happy to merge this PR for updating Windows Server OSes with the 32bit on 64bit values and if needed later on update the Windows Desktop OSes.

waynr:
:+1:

joshcooper:
> That made me wonder if we were going to be reworking our test matrices. And if in reworking them possibly including 32 bit puppet on 64 bit desktop OSes. And if that was a possibility, then would we want to include the same 32bit on 64bit hostgenerator option strings for desktop OSes that we're including for server OSes in the PR?

I want to deprecate, stop testing, and eventually remove support for 32-bit puppet on 64-bit windows. We currently test 32-bit puppet on 2012r2 and 2008r2 (both 64-bit), and this PR handles those combinations, so that we can at least support what we currently test. That said, I don't want to expand the set of 32/64 combinations to desktop platforms, since the goal is the opposite.

When we eventually remove 32-bit puppet from 64-bit windows, we will need to select a replacement 32-bit windows, and our choices are either 2008 non-R2 or desktop. Of those, we only have windows 10 32-bit in the vmpooler, so we will likely adopt that, but that will be 32-bit puppet on 32-bit windows.

PR #14, opened 2015-12-21
(QENG-3308)(QENG-3309) Add support for Cisco platforms

LuvCurves:


lib/beaker-hostgenerator/data/vmpooler.rb
```diff
@@ -35,6 +35,14 @@ module Vmpooler
           'platform' => 'el-7-x86_64',
           'template' => 'centos-7-x86_64'
         },
+        'cisco5-64' => {
+          'platform' => 'cisconxos-9-x86_64',
+          'template' => 'cisco-nxos-9k-x86_64'
```
waynr: Since the nxos template isn't on vmpooler yet, I'd prefer to postpone adding it here.

LuvCurves:
got it.

lib/beaker-hostgenerator/data/vmpooler.rb
waynr: Nevermind!

lib/beaker-hostgenerator/data/vmpooler.rb
```diff
@@ -35,6 +35,14 @@ module Vmpooler
           'platform' => 'el-7-x86_64',
           'template' => 'centos-7-x86_64'
         },
+        'cisco5-64' => {
+          'platform' => 'cisconxos-9-x86_64',
```
waynr: So what beaker does with this platform string is splits it into three components:
* platform
* version
* arch

The regex used to do this is here:
https://github.com/puppetlabs/beaker/blob/54b1801d9bb6187796142160d3c0828639c542d5/lib/beaker/platform.rb#L6

So it looks like Beaker wouldn't actually recognize this platform string. Also, the version string would then be misleading since this is actually `cisco 5`, not `cisco 7`.

My recommendation here would be to set the platform string to `cisco-5-x86_64`.

lib/beaker-hostgenerator/data/vmpooler.rb
```diff
@@ -35,6 +35,14 @@ module Vmpooler
...
+          'template' => 'cisco-nxos-9k-x86_64'
+        },
+        'cisco7-64' => {
+          'platform' => 'ciscoexr-9-x86_64',
```
waynr: Similar to my previous comment, my recommendation here would be to set the platform string to `cisco-7-x86_64`.

colinPL:
ok to test

waynr:
:+1:

lib/beaker-hostgenerator/data/vmpooler.rb
```diff
@@ -35,6 +35,14 @@ module Vmpooler
           'platform' => 'el-7-x86_64',
           'template' => 'centos-7-x86_64'
         },
+        'cisco5-64' => {
+          'platform' => 'cisco-5-x86_64',
+          'template' => 'cisco-nxos-9k-x86_64'
+        },
+        'cisco7-64' => {
```
joshcooper: In hipchat, @LuvCurves mentioned, "we have a production version of cisco-wrlinux-7 now. It is called 'cisco-exr-9k-x86_64'". Does that mean in the future we might have a different production version of `cisco-wrlinux-7` based on Nexus or IOS? If so, then I don't think `cisco7-64` is specific enough. In my mind the beaker-hostgenerator string should identify which vmpooler `template` to use, but not necessarily the `platform`. For example, `centos7-64` maps to the `centos-7-x86_64` template:

```
'centos7-64' => {
  'platform' => 'el-7-x86_64',
  'template' => 'centos-7-x86_64'
},
```

But both `centos7-64` and `redhat7-64` have platform `el-7-x86_64`.

waynr: @joshcooper I agree that we may need to be more specific in the platform string, but if that's the case then we will also have to update beaker here:
https://github.com/puppetlabs/beaker/blob/54b1801d9bb6187796142160d3c0828639c542d5/lib/beaker/platform.rb#L6

LuvCurves:
Wayne to take this over.

lib/beaker-hostgenerator/data/vmpooler.rb
waynr: @joshcooper After speaking with @LuvCurves and @shermdog about this, there is quite a bit of uncertainty about whether it will be necessary to distinguish between the various hardware platforms. My preference is to be more explicit so I have updated the PR to add a hardware platform distinction to the beaker-hostgenerator spec string.

joshcooper:
:+1:

PR #15, opened 2015-12-21
(QENG-3301) Add support for Cumulus Linux

LuvCurves:


lib/beaker-hostgenerator/data/vmpooler.rb
```diff
@@ -35,6 +35,10 @@ module Vmpooler
           'platform' => 'el-7-x86_64',
           'template' => 'centos-7-x86_64'
         },
+        'cumulus25-64' => {
+          'platform' => 'cumulusvx-25-x86_64',
```
joshcooper: should that be `cumulus-vx-...` like `template` uses?

LuvCurves: Current convention for platform seems to be osname-majorversionnumber-arch. @waynr, correct me if I'm wrong

waynr: @LuvCurves the convention you mention is correct, but Beaker only recognizes specific osnames: https://github.com/puppetlabs/beaker/blob/54b1801d9bb6187796142160d3c0828639c542d5/lib/beaker/platform.rb#L6

So this should probably be `cumulus-25-x86_64`

waynr:
ok to test

waynr:
:+1:

joshcooper:
`platform` version strings typically contain dotted version numbers, e.g. `osx-10.10-x86_64`, `ubuntu-15.04-amd64`, so I'm thinking cumulus should use`'platform' => 'cumulus-2.5-x86_64'`, assuming that's the correct OS version?

waynr:
@joshcooper I agree, if the version should actually be 2.5

LuvCurves:
@joshcooper That was IF the Cumulus version was 2.5, it's not. It's 25. So we can make it 25.0 if you'd like, but it's not 2.5

LuvCurves:
Updated platform string and name to match.  The actual version is 2.5.5

lib/beaker-hostgenerator/data/vmpooler.rb
```diff
@@ -35,6 +35,10 @@ module Vmpooler
           'platform' => 'el-7-x86_64',
           'template' => 'centos-7-x86_64'
         },
+        'cumulus2-64' => {
```
joshcooper: should this be `cumulus25-64` in case we need to support cumulus 2.6.x in the future?

LuvCurves:
Wayne taking over

lib/beaker-hostgenerator/data/vmpooler.rb
waynr: @joshcooper I went ahead and changed it to `cumulus25-64` since we don't know how versioning works for the Cumulus platforms and may need as you suggest to support 2.6 in the future.

joshcooper:
:+1:

PR #23, opened 2016-02-08
(QENG-3466) Fix --list bug and write tests to prevent recurrences.

waynr:


bin/beaker-hostgenerator
```diff
@@ -5,4 +5,8 @@ $LOAD_PATH.unshift(lib_dir)
 
 require 'beaker-hostgenerator'
 
-BeakerHostGenerator::CLI.new.execute!
+cli = BeakerHostGenerator::CLI.new
+
+if not cli then
```
doug-rosser: A cli instance is interpreted as a False value?

waynr: This is an alternative to calling `exit` directly within `BeakerHostGenerator::CLI.new` in the circumstance that `--list` is passed as an option on the command line; we return early from the `initialize` method with a `nil` return value. Since otherwise the `initialize` method returns a `BeakerHostGenerator::CLI` object, we can check for a "falsy" value here to determine whether or not to call `execute` (which `nil` does not have) on the returned object.

I'm actually not sure why returning early from the `initialize` method in Ruby causes the `new` method to return the object returned by `initialize` (`nil` in this case). My assumption is that `new` would allocate the object instance then call `initialize` to prepare that object. Maybe @rick or @kevpl could provide some insight here?

doug-rosser:
And other than my question above, :+1:

bin/beaker-hostgenerator
er0ck: should probably test the above in units...

CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ Versioning](http://semver.org).
 ## [Unreleased][unreleased]
 ### Changed
 
+## [0.4.0] - 2016-02-05
+- Add new optional parameters to CLI.
```
er0ck: such as?  what do they do?

bin/beaker-hostgenerator
waynr: Just a quick note, I had no idea what I was talking about up there--this particular change (which is undone in new commit) came about as a result of my misinterpretation of a stack trace I was seeing.

kevpl:
lots of maint commits, but no ticket ones?  Other than that, :+1:

PR #28, opened 2016-04-30
(QENG-3276) Add None hypervisor and per-host settings support

nwolfe:
This commit adds a new "none" hypervisor to support non-provisioned
hosts. Non-provisioned hosts require a handful of other host settings to
be specified in order for Beaker to use it properly, so this commit also
adds support for arbitrary key-value host settings.

For example:
```
 $ .. centos6-64m{hypervisor=none,hostname=my-host,vmhostname=othername}
 ---
 HOSTS:
   my-host:
     pe_dir:
     pe_ver:
     pe_upgrade_dir:
     pe_upgrade_ver:
     platform: el-6-x86_64
     hypervisor: none
     vmhostname: othername
     roles:
     - agent
     - master
 CONFIG:
   nfs_server: none
   consoleport: 443
```

### DIFF
--- a/lib/beaker-hostgenerator/cli.rb
+++ b/lib/beaker-hostgenerator/cli.rb
@@ -121,7 +121,7 @@ Usage: beaker-hostgenerator [options] <layout>
         raise BeakerHostGenerator::Exceptions::SafeEarlyExit
       else
         # Tokenizing the config definition for great justice
-        @tokens = argv[0].split('-')
+        @tokens = __tokenize_input(argv[0])
 
         if @options[:osinfo_version] === 0
           warning = <<-eow
@@ -135,6 +135,52 @@ eow
           STDERR.puts(warning)
         end
       end
+    end
+
+    # Breaks apart the host input string into chunks suitable for processing
+    # by the generator. Returns an array of substrings of the input spec string.
+    #
+    # The input string is expected to be properly formatted using the dash `-`
+    # character as a delimiter. Dashes may also be used within braces `{...}`,
+    # which are used to define arbitrary key-values on a node.
+    #
+    # @param spec [String] Well-formatted string specification of the hosts to
+    #                      generate. For example `"centos6-64m-debian8-32a"`.
+    # @returns [Array<String>] Input string split into substrings suitable for
+    #                          processing by the generator. For example
+    #                          `["centos6", "64m", "debian8", "32a"]`.
+    def __tokenize_input(spec)
+      # Here we allow dashes in certain parts of the spec string
+      # i.e. "centos6-64m{hostname=foo-bar}-debian8-32"
+      # by first replacing all occurrences of - with | that exist within
+      # the braces {...}.
+      #
+      # So we'd end up with:
+      #   "centos6-64m{hostname=foo|bar}-debian8-32"
+      #
+      # Which we can then simply split on - into:
+      #   ["centos6", "64{hostname=foo|bar}", "debian8", "32"]
+      #
+      # And then finally turn the | back into - now that we've
+      # properly decomposed the spec string:
+      #   ["centos6", "64{hostname=foo-bar}", "debian8", "32"]
+      #
+      # NOTE we've specifically chosen to use the pipe character |
+      # due to its unlikely occurrence in the user input string.
+      spec = String.new(spec) # Copy so we can replace characters inline
+      within_braces = false
+      spec.chars.each_with_index do |char, index|
+        case char
+        when '{'
+          within_braces = true
+        when '}'
+          within_braces = false
+        when '-'
+          spec[index] = '|' if within_braces
+        end
+      end
+      tokens = spec.split('-')
+      tokens.map { |t| t.gsub('|', '-') }
     end
 
     def print_platforms_and_roles
--- a/lib/beaker-hostgenerator/cli.rb
+++ b/lib/beaker-hostgenerator/cli.rb
@@ -1,4 +1,5 @@
 require 'beaker-hostgenerator/generator'
+require 'beaker-hostgenerator/hypervisor'
 require 'optparse'
 
 module BeakerHostGenerator
@@ -23,7 +24,7 @@ module BeakerHostGenerator
 Usage: beaker-hostgenerator [options] <layout>
 
  where <layout> takes the following form:
-  <platform>-<arch><roles>[[-<platform>]-<arch>[[<arbitrary-roles>,[...]].]<roles>[...]]
+  <platform>-<arch>[[<arbitrary-roles>,[...]].]<roles>[{<arbitrary-settings>,[...]}][-<arch>[[<arbitrary-roles>,[...]].]<roles>[{<arbitrary-settings>,[...]}]][-<layout>]
 
  examples:
   centos6-64mdca-32a
@@ -44,6 +45,11 @@ Usage: beaker-hostgenerator [options] <layout>
   centos6-32compile_master,another_role.ma
    1 CentOS 6 32 bit node with roles = master, agent, compile_master, another_role
 
+ example with arbitrary host settings:
+  centos6-64m{hypervisor=none\\,hostname=static1\\,my-key=my-value}-32a
+   1 CentOS 6 64 bit node with roles = master, hypervisor = none, node name = static1, and my-key = my-value
+   1 CentOS 6 32 bit node with roles = agent and the default hypervisor
+
  Generally, it is expected that beaker-hostgenerator output will be redirected to a file, for example:
   beaker-hostgenerator centos6-64ma > host.cfg
 
@@ -53,14 +59,14 @@ Usage: beaker-hostgenerator [options] <layout>
 
         opts.on('-l',
                 '--list',
-                'List beaker-hostgenerator supported platforms and roles. ' <<
+                'List beaker-hostgenerator supported platforms, roles, and hypervisors. ' <<
                 'Does not produce host config.') do
-          @options[:list_platforms_and_roles] = true
+          @options[:list_supported_values] = true
         end
 
         opts.on('-t',
                 '--hypervisor HYPERVISOR',
-                'Set beaker-hostgenerator hypervisor. ') do |h|
+                'Set beaker-hostgenerator default hypervisor. ') do |h|
           @options[:hypervisor] = h
         end
 
@@ -90,7 +96,7 @@ Usage: beaker-hostgenerator [options] <layout>
         end
 
         opts.on('--disable-default-role',
-                "Do not include the default /'agent/' role.") do
+                "Do not include the default 'agent' role.") do
           @options[:disable_default_role] = true
         end
 
@@ -116,8 +122,8 @@ Usage: beaker-hostgenerator [options] <layout>
 
       optparse.parse!(argv)
 
-      if @options[:list_platforms_and_roles]
-        print_platforms_and_roles
+      if @options[:list_supported_values]
+        print_supported_values
         raise BeakerHostGenerator::Exceptions::SafeEarlyExit
       else
         # Tokenizing the config definition for great justice
@@ -183,13 +189,14 @@ eow
       tokens.map { |t| t.gsub('|', '-') }
     end
 
-    def print_platforms_and_roles
-      puts "valid beaker-hostgenerator platforms:  "
+    # Prints to stdout a human-readable listing of all supported values for
+    # the following: platforms, architectures, roles, and hypervisors.
+    def print_supported_values
+      puts "valid beaker-hostgenerator platforms:"
       platforms = get_platforms(@options[:osinfo_version])
       platforms.each do |k|
         puts "   #{k}"
       end
-
       puts "\n"
 
       puts "valid beaker-hostgenerator architectures:"
@@ -198,9 +205,15 @@ eow
       puts "   6432 => 64-bit OS with 32-bit Puppet (Windows Only)"
       puts "\n"
 
-      puts "valid beaker-hostgenerator host roles:  "
+      puts "valid beaker-hostgenerator host roles:"
       ROLES.each do |k,v|
         puts "   #{k} => #{v}"
+      end
+      puts "\n"
+
+      puts "valid beaker-hostgenerator hypervisors:"
+      BeakerHostGenerator::Hypervisor.supported_hypervisors().keys.each do |k|
+        puts "   #{k}"
       end
     end
 
--- a/lib/beaker-hostgenerator/cli.rb
+++ b/lib/beaker-hostgenerator/cli.rb
@@ -1,5 +1,6 @@
 require 'beaker-hostgenerator/generator'
 require 'beaker-hostgenerator/hypervisor'
+require 'beaker-hostgenerator/roles'
 require 'optparse'
 
 module BeakerHostGenerator
@@ -169,7 +170,7 @@ eow
       puts "\n"
 
       puts "valid beaker-hostgenerator host roles:"
-      ROLES.each do |k,v|
+      BeakerHostGenerator::Roles::ROLES.each do |k,v|
         puts "   #{k} => #{v}"
       end
       puts "\n"
--- a/lib/beaker-hostgenerator/cli.rb
+++ b/lib/beaker-hostgenerator/cli.rb
@@ -126,8 +126,7 @@ Usage: beaker-hostgenerator [options] <layout>
         print_supported_values
         raise BeakerHostGenerator::Exceptions::SafeEarlyExit
       else
-        # Tokenizing the config definition for great justice
-        @tokens = tokenize_input(argv[0])
+        @layout = argv[0]
 
         if @options[:osinfo_version] === 0
           warning = <<-eow
@@ -144,7 +143,7 @@ eow
     end
 
     def execute
-      BeakerHostGenerator::Generator.new.generate(@tokens, @options)
+      BeakerHostGenerator::Generator.new.generate(@layout, @options)
     end
 
     def execute!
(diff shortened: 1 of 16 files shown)


lib/beaker-hostgenerator/hypervisor/vmpooler.rb
```diff
@@ -14,28 +14,18 @@ def global_config()
...
         # set hypervisor
-        host_config['hypervisor'] = 'vmpooler'
-
-        # generate node definition
-        ostype = node_info['ostype']
-        nodeid = node_info['nodeid']
-        bits = node_info['bits']
-
-        platform = "#{ostype}-#{bits}"
-        name = "#{platform}-#{nodeid}"
```
nwolfe: I moved the name generation out of the hypervisor and into the single generator because the name is no longer hypervisor-specific.

test/fixtures/per-host-settings/arbitrary-settings.yaml
```diff
@@ -0,0 +1,57 @@
+---
+arguments_string: redhat7-64role_a,role_b.m{hypervisor=none,hostname=redhat,vmhostname=vm-redhat,ip=1.2.3.4}-sles10-32a-64a{key=value-with-dashes}-64dc
```
nwolfe: This is a good holistic example of the changes in this PR

nwolfe:
Whoops, I forgot to update the `--help` output string in the CLI.

I'll add another commit for that on Monday.

nwolfe:
https://jenkins-qe.delivery.puppetlabs.net/job/qe_gem_unit-rb_rspec/

Wonderful, the `to_h` method is not available in Ruby < 2.  I'll refactor that on Monday as well.

nwolfe:
jenkins retest this please

lib/beaker-hostgenerator/cli.rb
```diff
@@ -121,7 +121,7 @@ def initialize(argv = ARGV.dup)
         raise BeakerHostGenerator::Exceptions::SafeEarlyExit
       else
         # Tokenizing the config definition for great justice
-        @tokens = argv[0].split('-')
+        @tokens = __tokenize_input(argv[0])
```
rick: Not a big fan of this naming convention for internal/private methods in ruby. I would either declare this `private` as `tokenize_input` or extract out to a class (seems like extraction is a a bit much to do for this small method, though).

lib/beaker-hostgenerator/generator.rb
```diff
@@ -104,6 +112,19 @@ def __parse_node_info_token(token)
         node_info['arbitrary_roles'] = []
       end
 
+      if node_info['host_settings']
```
rick: I would probably extract this code to a method with an intention-revealing name.

lib/beaker-hostgenerator/generator.rb
```diff
@@ -104,6 +112,19 @@ def __parse_node_info_token(token)
...
+          delete('{}').
+          split(',').
+          map { |keyvalue| keyvalue.split('=') }
+        ]
+      else
+        node_info['host_settings'] = {}
+      end
+
       return node_info
     end
```
rick: This is a pretty big method at this point; seems like it could be decomposed.

lib/beaker-hostgenerator/cli.rb
nwolfe: I don't either, was just sticking to the convention.  Although I notice that we didn't have the `__method` convention in this class, oops!  I'll remove that part and throw in a `private`

lib/beaker-hostgenerator/generator.rb
nwolfe: Good idea, will do

lib/beaker-hostgenerator/generator.rb
nwolfe: Did you intend to comment on the `generate` method instead?  https://github.com/puppetlabs/beaker-hostgenerator/pull/28/files#diff-0623c137fe4358c047889a4ff1adc01bR20

I only say that because I was thinking the same thing, and that method is longer and harder to read than this one.
But I'm happy to decompose both! 😄

lib/beaker-hostgenerator/cli.rb
waynr: I don't know if the convention is as important as understandability for rubystas, so +1 on whatever naming scheme makes most sense to folks with more experience in ruby.

lib/beaker-hostgenerator/generator.rb
waynr: My only request as far as decomposition goes is that we mark the new functions as private, otherwise decomposition ftw!

waynr:
Something I would be interested to see is a test fixture that exercises input that causes an exception state But since that would require writing spec test code to exercise the `expected_exception` key in the fixture hash, which might be unreasonable to ask, I won't withhold approval.

Generally speaking, I'm :+1: once all comments have been addressed.

lib/beaker-hostgenerator/generator.rb
nwolfe: Ugh, spec tests are all calling in to these methods directly so I can't mark them `private`.

I wish these spec tests didn't exist, and all testing was done via fixtures.

nwolfe:
@waynr @rick I added several new commits, which update the docs & CLI help text and continue with some further refactorings.

lib/beaker-hostgenerator/cli.rb
nwolfe: Moved into the new Parser module

lib/beaker-hostgenerator/generator.rb
nwolfe: I made several attempts to decompose this method and the larger `generate` method above, and I didn't like any of them.

I did change this code though, by moving it along with other tightly-coupled functions into the new Parser class.
I don't think decomposing `parse_node_info_token` any further would help much as it's not a deep method at all, just some datatype adjustments on a map.

lib/beaker-hostgenerator/generator.rb
```diff
@@ -32,43 +37,38 @@ def generate(tokens, options)
...
-
-        host_config['roles'].concat __generate_host_roles(node_info)
-        host_config['roles'].uniq!
-
-        if not options[:disable_role_config]
-          host_config['roles'].each do |role|
-            host_config.deep_merge! __get_role_config(role)
-          end
-        end
+        __generate_host_roles!(host_config, node_info, options)
```
nwolfe: Like I said in my other comment, I tried many times to decompose this cleanly, but I didn't like any of them.
There's too many local variables weaved through this method, and I believe it could be rearranged to allow for better decomposition, but I didn't want to do too much all at once.

I did extract all the role code into its own method below, so that helps a bit.

nwolfe:
This is ready for another round of review.

lib/beaker-hostgenerator/generator.rb
```diff
@@ -1,23 +1,28 @@
...
-    #               parameters provided on the command line.
-    def generate(tokens, options)
+    # @param layout [String] The raw hosts specification user input.
+    #                        For example `"centos6-64m-redhat7-64a"`.
+    # @param options [Hash] Global, optional configuration such as the default
+    #                       hypervisor or OS info version.
+    #
+    # @returns [String] A complete YAML map as a string defining the HOSTS and
+    #                   CONFIG sections as required by Beaker.
+    def generate(layout, options)
```
nwolfe: Like I said in my other comment, I tried many times to decompose this cleanly, but I didn't like any of them.
There's too many local variables weaved through this method, and I believe it could be rearranged to allow for better decomposition, but I didn't want to do too much all at once.

I did extract all the role code into its own method below, so that helps a bit.

lib/beaker-hostgenerator/generator.rb
rick: @nwolfe right -- I was thinking about `generate` and got a bit cross-eyed -- missed the method boundary.

This is definitely a good tidbit for ESO CodeCraft, if we wanted to pull the original version up on the screen and talk about ways to refactor / extract, and the tradeoffs between the tests we have, the tests we'd want (what those even might be), and what could / should be done to get there.

nwolfe:
> Something I would be interested to see is a test fixture that exercises input that causes an exception state But since that would require writing spec test code to exercise the expected_exception key in the fixture hash, which might be unreasonable to ask, I won't withhold approval.

@waynr

I went ahead and wired up support for the `expected_exception` here: https://github.com/puppetlabs/beaker-hostgenerator/pull/28/commits/78cccb144d471a678cdf5bc498037f49a0cec621

But then I ended up re-writing my failure tests directly as spec tests instead: https://github.com/puppetlabs/beaker-hostgenerator/pull/28/commits/b80d94ab85d3b9c44bd39eecc4622230898d9542

spec/beaker-hostgenerator/generator_spec.rb
```diff
@@ -98,12 +57,20 @@ module BeakerHostGenerator
             fixture_hash['environment_variables'].each do |key, value|
               ENV[key] = value
             end
+
             cli = BeakerHostGenerator::CLI.new(arguments)
-            test_hash = YAML.load(cli.execute)
+            if fixture_hash['expected_exception']
+              # Turn fully-qualified classname string into an actual Class object.
+              expected_class = eval fixture_hash['expected_exception']
```
nwolfe: `eval`!?!?!  Turns out Ruby doesn't have the most convenient way of turning a string classname into actual Class.  It doesn't support nesting of classes, so you have to walk the module/class hierarchy yourself, like this:
```ruby
        qualified_classname = fixture_hash['expected_exception'].split('::')
        expected_class = qualified_classname.reduce(Object) do |accum, e|
          accum.const_get(e)
        end
```

I opted to go with `eval` instead for simplicity sake.

nwolfe:
jenkins retest this please

lib/beaker-hostgenerator/generator.rb
nwolfe: Yeah, there's definitely room for improvement in this method.

I'd be happy to discuss this in ESO CodeCraft!

nwolfe:
This is polished and ready for merge.

lib/beaker-hostgenerator/generator.rb
```diff
@@ -32,43 +37,38 @@ def generate(tokens, options)
...
-                   hypervisor.generate_node(node_info, host_config, bhg_version)
+        host_config = hypervisor.generate_node(node_info, host_config, bhg_version)
         config['CONFIG'].deep_merge!(hypervisor.global_config())
 
+        # Merge in any arbitrary key-value host settings. Treat the 'hostname'
+        # setting specially, and don't merge it in as an arbitrary setting.
+        arbitrary_settings = node_info['host_settings']
+        if arbitrary_settings['hostname']
+          host_name = arbitrary_settings['hostname']
+          arbitrary_settings.delete('hostname')
```
rick: Trivial inlining:

``` ruby
  host_name = arbitrary_settings.delete('hostname') if arbitrary_settings.has_key?('hostname')
```

lib/beaker-hostgenerator/generator.rb
```diff
@@ -32,43 +37,38 @@ def generate(tokens, options)
...
+        end
+        host_config.merge!(arbitrary_settings)
```
rick: I can't tell clearly from this diff context, but is there any risk of direct changes to `host_config` impacting resources held by outside callers? (maybe whatever is being passed around by `base_host_config`)

If so, copying the hash early would get rid of any action-at-a-distance problems.

It's harder to reason about, but it might be safer to just copy the hash anyway, in case under maintenance a reuse which could be acted upon at a distance might get introduced at some point.

lib/beaker-hostgenerator/generator.rb
rick: Ditto earlier comment(s) about naming convention for "private" methods.

lib/beaker-hostgenerator/generator.rb
```diff
@@ -77,34 +77,21 @@ def generate(tokens, options)
...
-    end
-
-    def __parse_node_info_token(token)
-      node_info = NODE_REGEX.match(token)
-
-      if node_info
-        node_info = Hash[ node_info.names.zip( node_info.captures ) ]
+    def __generate_host_roles!(host_config, node_info, options)
+      if not options[:disable_default_role]
+        host_config['roles'] = ['agent']
```
rick: Similar questions here about action-at-a-distance when modifying passed-in hashes. Maybe copy and returned the result instead?

spec/beaker-hostgenerator/generator_spec.rb
rick: Yeah, the rails implementation is even hairier: https://github.com/rails/rails/blob/861b70e92f4a1fc0e465ffcf2ee62680519c8f6f/activesupport/lib/active_support/inflector/methods.rb#L232-L279

Eval is a pretty loaded gun. One thing we might do here is do a regex sanity check that there's nothing in the string but what could only be a class name.  Use `\A` and `\z` instead of `^` and `$` for this (due to multi-line security problems).

So maybe something like (double-check my regex, please):

``` ruby
  class_name = fixture_hash['expected_exception']
  expected_class = constantize(class_name) if class_name
end

def constantize(string)
  raise "Invalid class name '#{string}'" unless valid_class_name?(string)
  eval string
end

def valid_class_name?(string)
  string =~ %r{\A(?:[A-Z][A-Za-z0-9_]*)(?:::[A-Z][A-Za-z0-9_]*)*\z}
end
```

lib/beaker-hostgenerator/generator.rb
waynr: +1 on creating copies of hashes unless one intentionally means to create a reference. I wonder how often rubyists intend to copy an object with the assignment operator when what they are actually doing is creating a new reference to that object (`new_reference = old_reference`).

spec/beaker-hostgenerator/generator_spec.rb
nwolfe: > Eval is a pretty loaded gun. 

If we wanted to get rid of it the alternate implementation I provided above works fine as well.
`const_get` just doesn't know how to recurse into the each module, so you have to hold it's hand.

lib/beaker-hostgenerator/generator.rb
nwolfe: Rather than having callers create copies of hashes, I've defined the hashes as methods that return a hash rather than a const that is a hash: https://github.com/nwolfe/beaker-hostgenerator/blob/qeng3276/none-hypervisor-and-host-settings/lib/beaker-hostgenerator/data.rb#L56-L63

nwolfe: > If so, copying the hash early would get rid of any action-at-a-distance problems.

So to clarify my above comment, I've copied at the earliest stage possible.

lib/beaker-hostgenerator/generator.rb
nwolfe: I wouldn't say this is action at a distance since this is a private function that lives below the only other function it's used in. It's a completely encapsulated function that sets up the host role portion of the data structure.

I suppose we could take the approach of:
```ruby
host_config = __generate_host_roles(host_config, node_info, options)
```
But then `__generate_host_roles` would have to clone the object.

Given that these are all private functions that are meant to decompose the larger `generate` method, I think we can take the liberty of writing the private functions in a way that is most convenient for the Generator and the `generate` method.

I actually was thinking that we'd end up with several functions of the form `configure_something_about_the_host_config!(host_config, ...)`, and then the `generate` method would just call them like so:
```ruby
def generate(...)
  host_config = base_host_config(options)
  configure_X!(host_config, options)
  configure_Y!(host_config, options)
  configure_Z!(host_config, options)
  host_config
end
```

lib/beaker-hostgenerator/generator.rb
nwolfe: Yeah I can make this one function private, but I can't make the existing `__generate_host_roles` (just realized my naming collision here) one private because of our spec tests, which directly call it.

spec/beaker-hostgenerator/generator_spec.rb
nwolfe: I'll change this to use my alternate implementation, and add a simple test fixture to exercise it.

lib/beaker-hostgenerator/generator.rb
nwolfe: Ah, yes

spec/beaker-hostgenerator/generator_spec.rb
nwolfe: Updated this here: https://github.com/puppetlabs/beaker-hostgenerator/pull/28/commits/b0926b2216f419bbd56ddbed71dc418efeaa822f#diff-8bdb9a6d73949911cc906d87d7f4aef8R67

lib/beaker-hostgenerator/generator.rb
nwolfe: Made it private here: https://github.com/puppetlabs/beaker-hostgenerator/pull/28/commits/b0926b2216f419bbd56ddbed71dc418efeaa822f#diff-0623c137fe4358c047889a4ff1adc01bR92

lib/beaker-hostgenerator/generator.rb
nwolfe: I realized the naming collision I introduced here (we had both `__generate_host_roles` and `__generate_host_roles!`) so I renamed one to `get_host_roles`.

I'm going to need more convincing if we want to add a bunch of hash cloning in these methods 😄

nwolfe:
Feedback addressed with another commit and some comments

waynr:
:+1:

nwolfe:
@rick Any more thoughts?

I'm happy to continue working on this code, but this PR/ticket has been dragging for a while now and it would be nice to get it in in this sprint (by tomorrow).  
If there's additional changes we'd like to make in here, I'd love to make them, and I might make them in my spare time, or otherwise lump those changes in with the next ticket that brings us back here.

rick:
It's fine -- everything I had was more advisory (and even the `eval`, I realize now was in tests not in the cmdline tool, which has less of a footprint), and nothing blocking.

:+1:

PR #31, opened 2016-05-20
(QENG-3377) Add --version CLI flag

nwolfe:
This commit adds a -v/--version CLI flag to print the version number.

This also refactors the flow-of-control to remove the need for the
SafeEarlyExit exception by moving all output-generating codepaths out of
the initialize method and into the execute method, and teasing apart the
--list and default generation codepaths. The execute method
now chooses the code path based on the CLI options set in initialize.

While we no longer throw a SafeEarlyExit exception, it's effectively
part of the API so we must wait until the appropriate time/release to
delete it. Existing usages of it should not be broken with the changes
in this commit.

### DIFF
--- a/lib/beaker-hostgenerator/cli.rb
+++ b/lib/beaker-hostgenerator/cli.rb
@@ -1,6 +1,7 @@
 require 'beaker-hostgenerator/generator'
 require 'beaker-hostgenerator/hypervisor'
 require 'beaker-hostgenerator/roles'
+require 'beaker-hostgenerator/version'
 require 'optparse'
 
 module BeakerHostGenerator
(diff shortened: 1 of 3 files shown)


lib/beaker-hostgenerator/cli.rb
```diff
@@ -116,69 +117,81 @@ def initialize(argv = ARGV.dup)
...
         end
       end
 
       optparse.parse!(argv)
+      @layout = argv[0]
+    end
 
-      if @options[:list_supported_values]
-        print_supported_values
-        raise BeakerHostGenerator::Exceptions::SafeEarlyExit
```
nwolfe: @waynr We don't need the SafeEarlyExit anymore with these changes, so we could delete the class, but...

It's part of our API, and since we're holding off on breaking changes until 1.0 I left it defined.
I'll add a ticket to delete it to that Epic.

nwolfe:
@waynr I'd like to include these changes in the next release too.

lib/beaker-hostgenerator/cli.rb
waynr: I think it's a relatively low-risk member of the API. That coupled with the fact that the `0` major version is intended to be unstable should give us the leeway to remove it now if you prefer. I don't have a preference in this case as I did in the past because in the past the API changes had a higher risk of breaking actual API use cases.

waynr:
:100: looks good to me!

kevpl:
👍

PR #32, opened 2016-05-21
(QENG-3808) Support uppercase alphanumeric architectures

nwolfe:
This commit adds support for uppercase-only alphanumeric architecture
bits in platform names. For example, we now support "POWER" or "POWER7"
instead of just "32" or "64".

Uppercase letters allow for backwards compatibility when roles are
specified, which come after the architecture bit. This implementation
relies on the fact that the role and arbitrary role characters are
only lowercase.

This allows us to distinguish between the architecture "POWER" and the
role "m" when parsing the spec "POWERm", as in "aix71-POWERm".

### DIFF
--- a/lib/beaker-hostgenerator/parser.rb
+++ b/lib/beaker-hostgenerator/parser.rb
@@ -14,7 +14,28 @@ module BeakerHostGenerator
   # each token you would call `is_ostype_token?` and/or `parse_node_info_token`.
   module Parser
 
-    # Capture role and bit width information about the node.
+    # Parses a single node definition into the following components:
+    #
+    #   * bits             Uppercase-only alphanumeric
+    #                      Examples: 32, 64, POWER, S390X
+    #
+    #   * arbitrary_roles  Lowercase-only alphabetical & underscores
+    #                      Examples: myrole, role_a
+    #
+    #   * roles            Lowercase-only, any combination of: u, a, c, l, d, f, m
+    #                      Examples: m, mdca
+    #
+    #   * host_settings    Any character (see `settings_string_to_map` for details)
+    #                      Examples: {hostname=foo-bar, ip.address=123.4.5.6}
+    #
+    # This regex is the main workhorse for parsing input to beaker-hostgenerator.
+    # There is a bit of pre and post parsing that happens before and after this
+    # regex though.
+    # Before we use this regex, we split the input containing multiple node
+    # definitions into individual node definitions to be parsed by this regex
+    # (see `tokenize_layout`).
+    # After we use this regex, we turn the host_settings key-value string into
+    # a proper Hash map (see `settings_string_to_map`).
     #
     # See Ruby Regexp class for information on the capture groups used below.
     # http://ruby-doc.org/core-2.2.0/Regexp.html#class-Regexp-label-Character+Classes
@@ -40,7 +61,7 @@ module BeakerHostGenerator
     #   * agent
     #   * database
     #
-    NODE_REGEX=/\A(?<bits>\d+)((?<arbitrary_roles>([[:lower:]_]*|\,)*)\.)?(?<roles>[uacldfm]*)(?<host_settings>\{[[:print:]]*\})?\Z/
+    NODE_REGEX=/\A(?<bits>[A-Z0-9]+|\d+)((?<arbitrary_roles>([[:lower:]_]*|\,)*)\.)?(?<roles>[uacldfm]*)(?<host_settings>\{[[:print:]]*\})?\Z/
 
     module_function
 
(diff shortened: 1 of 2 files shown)


nwolfe:
@waynr What do you think of adding UPPERCASE-only alphanumeric support?  That way we don't have to wait until 1.0 to introduce a breaking change and remove the case-sensitivity.

We'll be able to add the AIX and SPARC platforms with these changes, something like `aix71-POWER` or `aix71-POWER6`.  At that point we'd be able to start getting rid of nearly all the puppet/agent/facter/hiera/etc static host files.

@johnduarte Do you think this would suffice for https://tickets.puppetlabs.com/browse/QENG-3808 until we can do a 1.0 release and implement it as you suggest in that ticket?

kevpl:
seems fine to me. This doesn't need to gate this PR in any way, but it would be nice if your description in the PR about the the possible architecture values was in some sort of help text or doc.

waynr:
@nwolfe @kevpl I would prefer that this PR also include developer documentation for valid architecture values when adding new platforms. Otherwise, I'm :+1: on the approach of using uppercase but would like @johnduarte to have a chance to acknowledge this is an acceptable fit for his needs.

nwolfe:
> but it would be nice if your description in the PR about the the possible architecture values was in some sort of help text or doc.

> I would prefer that this PR also include developer documentation for valid architecture values when adding new platforms

Yep, this PR started out as a proof-of-concept that I put up for discussion. I'll add documentation for this.

lib/beaker-hostgenerator/parser.rb
```diff
@@ -14,7 +14,28 @@ module BeakerHostGenerator
...
+    #                      Examples: {hostname=foo-bar, ip.address=123.4.5.6}
+    #
+    # This regex is the main workhorse for parsing input to beaker-hostgenerator.
+    # There is a bit of pre and post parsing that happens before and after this
+    # regex though.
+    # Before we use this regex, we split the input containing multiple node
+    # definitions into individual node definitions to be parsed by this regex
+    # (see `tokenize_layout`).
+    # After we use this regex, we turn the host_settings key-value string into
+    # a proper Hash map (see `settings_string_to_map`).
```
nwolfe: @waynr @kevpl Added some docs explaining the data requirements.

I also think it'll help once I add a platform that uses the UPPERCASE support, like I'm about to do for `aix71-POWER` et al

nwolfe:
Added some developer notes/docs.

Ready for more feedback / merge

lib/beaker-hostgenerator/parser.rb
```diff
@@ -14,7 +14,28 @@ module BeakerHostGenerator
   # each token you would call `is_ostype_token?` and/or `parse_node_info_token`.
   module Parser
 
-    # Capture role and bit width information about the node.
+    # Parses a single node definition into the following components:
+    #
+    #   * bits             Uppercase-only alphanumeric
+    #                      Examples: 32, 64, POWER7 (note 7POWER is invalid)
```
johnduarte: It seems unintuitive to me that we would support `POWER7` but not `7POWER`.

This particular pattern is a bit restrictive. We already have a use case that does not fit into it with the `S390X` architecture for zLinux.

I applaud the backward compatible solution, but would prefer to see a regex pattern that grabs everything before the first lower case alpha character as the given architecture.

lib/beaker-hostgenerator/parser.rb
```diff
@@ -14,7 +14,28 @@ module BeakerHostGenerator
...
+    #
+    #   * arbitrary_roles  Lowercase-only alphabetical & underscores
+    #                      Examples: myrole, role_a
```
johnduarte: This is freakin' awesome. Thanks @nwolfe

lib/beaker-hostgenerator/parser.rb
nwolfe: Ah, `S390X` wouldn't work with this, you're right.  I don't think there's any reason we can't improve this as you suggest, I'll make that happen.

Glad you caught this.

nwolfe: Done in https://github.com/puppetlabs/beaker-hostgenerator/pull/32/commits/1ecf4d1135662cc44b39a749e044fa6a7e2d4413

nwolfe:
@johnduarte @waynr Ok I've added support for mixed alphanumerics like `S390X`. I also added a few more spec tests that validate parsing with roles specified.

This should be ready for merge now.

johnduarte:
:+1:

johnduarte:
@nwolfe do we need others to sign off on this, or would you like me to merge?

waynr:
:+1:

PR #34, opened 2016-06-04
(QENG-3946) Support arbitrary hypervisors

nwolfe:
This commit allows unknown hypervisors to be specified by the user.
Previously, if the hypervisor specified by the user was not one of the
valid, built-in implementations then an error was thrown notifying the
user that an invalid hypervisor was requested.
Now, this is no longer an error, and instead the host generation will
continue as normal but there will be no hypervisor-specific
configuration. The name of the unknown hypervisor will be the value of
the hypervisor configuration of the hosts.

To implement this, we can repurpose the special `Hypervisor::None`
implementation to instead represent an unknown (but equally valid)
hypervisor. The unknown name will now be the value of `hypervisor` in
the generated config, where it used to be `hypervisor: none`.

For example:
```bash
  $ .. centos6-64{hypervisor=custom}
```
```yaml
  ---
  HOSTS:
    centos6-64-1:
      pe_dir:
      pe_ver:
      pe_upgrade_dir:
      pe_upgrade_ver:
      platform: el-6-x86_64
      hypervisor: custom
      roles:
      - agent
  CONFIG:
    nfs_server: none
    consoleport: 443
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,19 +3,17 @@
 `beaker-hostgenerator` is a command line utility designed to generate beaker
 host config files using a compact command line SUT specification.
 
-It currently supports Puppets' internal [vmpooler][vmpooler] hypervisor and
-static (non-provisioned) nodes, and is designed in a way that makes it possible
-to easily add support for additional hypervisors (any hypervisor type supported
-by [beaker][beaker]).
-
 <!-- markdown-toc start - Don't edit this section. Run M-x markdown-toc-generate-toc again -->
 **Table of Contents**
 
 - [Beaker Host Generator](#beaker-host-generator)
+    - [Hypervisors](#hypervisors)
     - [Usage](#usage)
         - [Simple two-host layout](#simple-two-host-layout)
         - [Single host with Arbitrary Roles](#single-host-with-arbitrary-roles)
         - [Two hosts with multiple hypervisors and arbitrary host settings](#two-hosts-with-multiple-hypervisors-and-arbitrary-host-settings)
+        - [Arbitrary global configuration settings](#arbitrary-global-configuration-settings)
+        - [Custom hypervisor](#custom-hypervisor)
     - [Testing](#testing)
         - [Test Fixtures](#test-fixtures)
             - [Generated Fixtures](#generated-fixtures)
@@ -23,6 +21,23 @@ by [beaker][beaker]).
     - [License](#license)
 
 <!-- markdown-toc end -->
+
+## Hypervisors
+
+Any hypervisor may be specified and generated, but if it's not a built-in
+hypervisor you will have to provide the entire hypervisor configuration as
+input. See the [Custom hypervisor](#custom-hypervisor) example for more
+information.
+
+It currently provides built-in configuration for Puppets' internal
+[vmpooler][vmpooler] hypervisor and static (non-provisioned) nodes, and is
+designed in a way that makes it possible to easily add support for additional
+hypervisors (any hypervisor type supported by [beaker][beaker]).
+
+To see the list of built-in hypervisors you can run:
+```
+$ beaker-hostgenerator --list
+```
 
 ## Usage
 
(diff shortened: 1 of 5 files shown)


nwolfe:
Ping @waynr @johnduarte for review

johnduarte:
@nwolfe this PR looks good to me. :+1: 

My only concern is that we are obscuring the 'none' hypervisor from being a 'valid' hypervisor from Beaker's point of view. As a nieve user of beaker-hostgenerator the docs convey that 'vmpooler' is the only 'valid' hypervisor, but I can then enter an arbitrary one. It is clear that I can do that either individually or globally via the '-t' flag. I just want to make sure from a UX standpoint the user can understand what a 'valid' hypervisor means going forward.

nwolfe:
> My only concern is that we are obscuring the 'none' hypervisor from being a 'valid' hypervisor from Beaker's point of view. As a nieve user of beaker-hostgenerator the docs convey that 'vmpooler' is the only 'valid' hypervisor

Ah yes, I renamed this internally from being "valid" or "supported" hypervisors to instead be "built-in" hypervisors.  The implication is that a built-in hypervisor has some extra configuration that will be generated when used.  

I'll adjust the README and CLI help text some more to try and convey this meaning.

I think we're slowly moving toward embedding beaker-hostgenerator within beaker itself, and then expanding beaker's `--config` argument to be the beaker-hostgenerator string, like `centos6-64mdca` or whatever.  The point is that the user will no longer be aware that beaker-hostgenerator is a standalone library, so this sort of mismatch between the two tools will go away.

johnduarte:
>Ah yes, I renamed this internally from being "valid" or "supported" hypervisors to instead be "built-in" hypervisors. The implication is that a built-in hypervisor has some extra configuration that will be generated when used. 

The "built-in" name makes perfect sense to me. Thanks @nwolfe !

waynr:
:+1: looks good to me!

johnduarte:
@nwolfe are you going to roll the documentation updates into this PR, or would you prefer that it get merged as is?

nwolfe:
> are you going to roll the documentation updates into this PR, or would you prefer that it get merged as is?

I'm going to add them to this PR.  Thank you for your attention!  Honestly 🔆

README.md
```diff
@@ -24,6 +22,23 @@ by [beaker][beaker]).
 
 <!-- markdown-toc end -->
 
+## Hypervisors
```
nwolfe: @johnduarte Added this section to the README, and example below

nwolfe:
> I'm going to add them to this PR. Thank you for your attention! Honestly 🔆

Done and squashed in.  I updated the README and changed the verbiage in the `--list` output to say `built-in hypervisors` instead of `valid hypervisors`.

nwolfe:
This is ready for review / merge

README.md
```diff
@@ -171,6 +186,39 @@ CONFIG:
...
+
+The following example shows one way of generating a custom hypervisor that
+includes both per-host configuration and global configuration.
+
+The term "custom" in this case signifies that it's not a built-in hypervisor
+(like `vmpooler`), which means we'll have to provide all the configuration
+ourselves as there isn't any built-in configuration for our hypervisor.
+
+```
+$ beaker-hostgenerator --hypervisor=custom --global={custom_api=http://api.custom.net} centos6-64{hypervisor=custom}
```
johnduarte: Defining `custom` with both a global flag and a host specific hash may lead users to infer that both are required. Should we omit the global flag or set the global flag to a different value to illustrate the disparate effect of these two entries?

nwolfe: They are both required.  

The `--global` is to provide the `custom_api` value in the `CONFIG` section, so you don't have to specify it on every single host.  This is akin to the `pooler_api` value we generated when specifying a `vmpooler` hypervisor.

The host-specific setting is to get the `hypervisor: custom` value on the hosts, which is necessary since hypervisors are host-specific.

johnduarte: Sorry, I was not referring to the `--global` flag, but the `--hypervisor=custom` flag as the *global flag* in my comment above.

nwolfe: Ahh whoops, good catch, you're right.  I started writing this example thinking I'd have 2 hosts specified, but didn't end up doing that.  I'll adjust this, good 👀

nwolfe: Updated. I removed the host setting and left the `--global` setting.

johnduarte:
:+1:  LGTM @nwolfe

PR #35, opened 2016-06-08
(MAINT) Sort platforms alphabetically

nwolfe:
Just so it's easier to navigate the large structure in the code.

----

Since we sorted the platforms alphabetically, the fixture generation
results in different filenames. Basically every time we modify
Data#platforms() we should regenerate the fixtures so we have all
platforms covered.

This moves the generated fixtures into a new directory called
`test/fixtures/generated/` so it's easier to work with them and not have
the manually created fixtures get in the way.


nwolfe:
@waynr

kevpl:
👍 to the first commit.

The second commit is a little weird. It's comparing a lot of files that have small changes like they change roles. Is this just git selecting bad files to compare?

nwolfe:
> should these files be having these small changes in them?

Yes, the small changes are a consequence of how our fixture generation code works.  I regenerated the fixtures because I sorted the platform map alphabetically, and the fixture generation is tied to the order of elements in the map + host roles.

Since we changed the order of elements in the map, fixture generation ends up creating slightly different pairs of <platform>+<role>, which is why those small changes you're referring to are mostly/all just different roles on existing fixtures.

I tried making the generation not so fragile and more deterministic, but it's not trivial, and if I'm going to spend that much time on them I'd rather refactor how/when/why/what we generate fixtures for (rather than continue with this every possible combination approach).

kevpl:
ok, cool. Thanks for the explanation @nwolfe! I'm 👍 overall.

PR #42, opened 2016-06-29
(MAINT) Regenerate test fixtures for new platforms

nwolfe:
I blew away the whole fixtures/generated/ directory and regenerated them
all again, so the new platforms will be included and tested.


smcelmurry:
👍

ferglor:
👍

PR #43, opened 2016-07-13
(QENG-4034) Automatically URL-decode raw input

nwolfe:
This commit adds a "pre-processing" step to the Parser that will always
attempt to URL-decode the input before parsing it. This behavior is
necessary to support specifying arbitrary settings in the input string
when the input ends up as part of an HTTP URL, such as almost every
usage we currently have in Jenkins.

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,6 +219,59 @@ CONFIG:
   custom_api: http://api.custom.net
 ```
 
+### URL-encoded input
+
+It may be necessary to URL-encode the input in order for it to properly be used
+in certain contexts, such as Jenkins.
+
+In most cases it will only be necessary to escape the characters that support
+arbitrary settings, which means the following three characters: {,}
+
+```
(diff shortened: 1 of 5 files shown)


nwolfe:
@waynr @johnduarte For review

I'm confident these are the changes necessary, but I still want to verify we can use these URL-encoded inputs as `test_targets` in Jenkins so I'm going to create a simple job with an axis to reproduce the buggy behavior and validate these changes fix it.

johnduarte:
:+1:  LGTM. Thanks @nwolfe

waynr:
:+1:  looks good to me also, let me know when you've had a chance to try it out on a Jenkins instance.

nwolfe:
@waynr @johnduarte Jenkins job demonstrating this will work: https://jenkins-qe.delivery.puppetlabs.net/job/beaker-hostgenerator_layout_axis_test/

PR #48, opened 2016-08-18
(QENG-4204) Add AlwaysBeScheduling hypervisor

nwolfe:
This commit adds a new hypervisor called 'abs' that will support
generating configurations necessary for working with CI.next and the
AlwaysBeScheduling service.

Currently, the behavior of the abs hypervisor is basically the same as
the vmpooler hypervisor. The only difference is it doesn't fixup the
nodes to support old PE versions.

Having a built-in hypervisor for ABS alleviates the need for users to
manually override the hypervisor key on each node.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,9 +31,10 @@ input. See the [Custom hypervisor](#custom-hypervisor) example for more
 information.
 
 It currently provides built-in configuration for Puppets' internal
-[vmpooler][vmpooler] hypervisor and static (non-provisioned) nodes, and is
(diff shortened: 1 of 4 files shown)


nwolfe:
@puppetlabs/cinext for review

waynr:
This should probably come with documentation explaining what "ABS" is, otherwise it looks pretty good to me.

nwolfe:
@waynr We don't really say much about the hypervisors as it is, but I added mention of it in the README (with a link to the private repo) and some docstrings.

PR #51, opened 2016-08-26
(QENG-4204) Add CLI flag to only generate templates

nwolfe:
This adds a CLI switch '--templates-only' that will reduce the generated
output to include only the template values from the HOSTS.


nwolfe:
@joshcooper @waynr For review.

We've been rolling with these changes in the CI.next branch for a week now.

PR #54, opened 2016-09-23
(QENG-4407) Return proper JSON map

nwolfe:
Previously the --templates-only switch would output a custom CSV-style
format, plus it repeated platforms rather than tallying them up.

This commit changes the output format to be a JSON map, with the
platforms as keys and the total counts as values.

### DIFF
--- a/lib/beaker-hostgenerator/abs-support.rb
+++ b/lib/beaker-hostgenerator/abs-support.rb
@@ -3,13 +3,28 @@ module BeakerHostGenerator
   module AbsSupport
     module_function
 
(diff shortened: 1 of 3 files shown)


nwolfe:
@joshcooper @rick For review.  If we're not ready for these changes, or if we're taking this CI.next support out of BHG, I'm fine with that.  This was easy enough to just do, I won't be hurt if we scrap it 😄

nwolfe:
Reminder to apply this change to the `cinext` branch if accepted.

PR #57, opened 2016-10-01
(BKR-934) Allow for whitespace in settings

nwolfe:
Previously all whitespace was automatically removed from all input. This
commit removes that behavior, allowing for values with whitespace in
them to be properly generated.

This change means users are responsible for properly quoting, escaping,
or URL-encoding the whitespace as necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,11 +227,12 @@ It may be necessary to URL-encode the input in order for it to properly be used
 in certain contexts, such as Jenkins.
 
 In most cases it will only be necessary to escape the characters that support
-arbitrary settings, which means the following three characters:
+arbitrary settings, which means the following four characters:
 
 - `{` is `%7B`
 - `,` is `%2C`
 - `}` is `%7D`
+- ` ` is `%20`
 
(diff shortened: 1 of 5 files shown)


nwolfe:
@waynr @kevpl For review.

nwolfe:
I suppose this is technically a breaking change, so once this is merged I'll do an `0.8.0` release.

kevpl:
I'm 👍 on the change. I'm confused about the version number though. Breaking changes are Y releases? Do we have a link/doc explaining why they're note X releases?

waynr:
@kevpl we're following the semver rule for `0.x` versions: http://semver.org/#spec-item-4

Basically in major version `0` anything can change at any time (so API is unstable). But for the sake of API consistency and loose pinning in project repos we are staying backwards compatible API-wise on patch level versions and bumping minor versions for small backwards incompatible changes.

nwolfe:
@kevpl Yep, what Wayne said.

@waynr Do you approve of these changes?

kevpl:
ah cool. Thanks for the clarification guys!

waynr:
chucknorris approves

PR #60, opened 2016-12-02
(MAINT) Exclude test files from built gem

nwolfe:
This project has a TON of test fixtures that are not necessary in production,
yet we were still bundling them into the gem.

We don't need to include any of these, they were just making the resulting gem
significantly larger.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ junit
 acceptance-tests
 pkg
 Gemfile.lock
+Gemfile.local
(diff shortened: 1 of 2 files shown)


nwolfe:
@waynr @joshcooper For review.

I noticed this when working on the puppet-agent pipeline and inspecting the various component `.tar.gz` files.  There's no reason for these 500+ files to dwarf the 10 source files in the gem.

PR #61, opened 2016-12-12
(MAINT) Delete test_files directive to keep fixtures out of the gem

nwolfe:
This gemspec directive `test_files` will cause the files it references to be
included in the built gem so they can be used during the `gem test` command.

We do not intend to support the ability for users to run `gem test`, and
furthermore this gem subcommand doesn't even exist anymore.


nwolfe:
@waynr @joshcooper This time, I actually built and unpacked the gem to verify the test files are not in the gem.

Turns out `test_files` is deprecated / no longer used.
Some references:
* http://stackoverflow.com/questions/18871541/what-is-the-purpose-of-test-files-configuration-in-a-gemspec
* https://github.com/rubygems/rubygems/issues/735

PR #62, opened 2016-12-13
(MAINT) Remove acceptance testing

nwolfe:
This deletes everything related to acceptance testing.

We had no tests, and we have no intentions of adding any. We started to lay out
the groundwork for acceptance testing a while ago with the "acceptance/"
directory, `beaker` test dependency, and the `acceptance` Rake task.

This commit deletes that groundwork, as it is out of date and just getting in
the way at this point.  If we decide to add acceptance tests in the future, it
will likely look different than this anyway.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,12 +2,6 @@ source ENV['GEM_SOURCE'] || "https://rubygems.org"
 
 gemspec
 
-
-group :testing do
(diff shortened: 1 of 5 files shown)


joshcooper:
Want to rebase?

nwolfe:
Rebased

PR #66, opened 2017-01-03
Add Ubuntu 16.10 platforms and test fixtures

ScottGarman:
Added Ubuntu 16.10 platforms for i386 and x86_64. The test fixtures were created by running rake generate:fixtures.

### DIFF
--- a/test/fixtures/generated/default/ubuntu1610-32l
+++ b/test/fixtures/generated/default/ubuntu1610-32l
@@ -1,0 +1,21 @@
+---
+arguments_string: ubuntu1610-32l
+environment_variables: {}
+expected_hash:
+  HOSTS:
+    ubuntu1610-32-1:
+      pe_dir: 
+      pe_ver: 
+      pe_upgrade_dir: 
+      pe_upgrade_ver: 
+      hypervisor: vmpooler
+      platform: ubuntu-16.10-i386
+      template: ubuntu-1610-i386
+      roles:
+      - agent
+      - classifier
+  CONFIG:
+    nfs_server: none
+    consoleport: 443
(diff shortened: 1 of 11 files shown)


test/fixtures/generated/default/ubuntu1610-32l
```diff
@@ -0,0 +1,21 @@
...
+      platform: ubuntu-16.10-i386
+      template: ubuntu-1610-i386
+      roles:
+      - agent
+      - classifier
+  CONFIG:
+    nfs_server: none
+    consoleport: 443
+    pooling_api: http://vmpooler.delivery.puppetlabs.net/
+expected_exception:
```
kurtwall: Is this boilerplate or is an entry for 32-bit systems necessary? I didn't realize we were still testing 32-bit *NIX platforms.

ScottGarman: To my understanding, we've dropped support for 32-bit puppet servers, but where the puppet-agent is concerned, we are still supporting and testing 32-bit platforms.

McdonaldSeanp: +1, that is also my undestanding @ScottGarman

McdonaldSeanp:
👍 LGTM

geoffnichols:
I think this looks good, but pinging @nwolfe for review.

johnduarte:
I realize this is after the fact, but Ubuntu 16.10 was classified as a Tier 3 platform and is not expected to receive coverage in CI. As such, this addition not necessary.

Ref: https://tickets.puppetlabs.com/browse/QA-2734

PR #68, opened 2017-01-24
(PA-893) Add AIX 7.2

branan:


geoffnichols:
Ping @nwolfe and @johnduarte for review.

johnduarte:
These changes are consistent with aix7.1.
I do not know what the significance of the 'l' suffix is at the end of the generated osinfo files. I see that the other aix variant have unique suffixes. Should the suffix for 7.1 and 7.2 be the same or should the be unique?

Listing of generated aix osinfo-version0
```
ls -1 test/fixtures/generated/osinfo-version-0/aix*
test/fixtures/generated/osinfo-version-0/aix53-POWERa
test/fixtures/generated/osinfo-version-0/aix61-POWERu
test/fixtures/generated/osinfo-version-0/aix71-POWERl
test/fixtures/generated/osinfo-version-0/aix72-POWERl
```

branan:
@johnduarte presumably they're just testing different PE roles (I know "a" is agent, and "l" is console). I'm not sure if there is an intended pattern to how new hosts exercise various roles in fixture generation. I  just copied 7.1 and sed'd it

johnduarte:
Sounds reasonable @branan .
:+1: 
@nwolfe since you originally created the generated files, does correspond with your expectations?

nwolfe:
@branan FYI all the fixtures under the `generated/` directory come from running the rake task, with some manual steps after.

The workflow we have when adding a new platform is this:

1. Add the platform to `data.rb`
2. Run rake task to generate fixtures (like 500+ of them)
3. Git commit only the fixtures that have the new platform in them (and delete/don't-commit the other created/changed fixtures)

The behavior during generation is randomly combine every platform with some random combination of roles and switches.

It's fine to have some created by hand in there; no need to change the fixtures you've got here.

PR #69, opened 2017-01-30
(BKR-1032) Switched from using STDERR to using $stderr

ferglor:
Using $stderr permits redirecting of stderr to a log file or buffer
other than the default STDERR

### DIFF
--- a/bin/genconfig2
+++ b/bin/genconfig2
@@ -5,7 +5,7 @@ $LOAD_PATH.unshift(lib_dir)
 
 require 'beaker-hostgenerator'
 
-STDERR.puts("Warning: 'genconfig2' is deprecated and will be removed in beaker-hostgenerator 1.x\n\n")
+$stderr.puts("Warning: 'genconfig2' is deprecated and will be removed in beaker-hostgenerator 1.x\n\n")
 
 begin
   cli = BeakerHostGenerator::CLI.new
--- a/test/test_stdout.rb
+++ b/test/test_stdout.rb
@@ -18,4 +18,9 @@ class TestStdout < Minitest::Test
       assert_match(/.*#{spec_key}.*/, @stdout.string)
     end
   end
+
+  def test_ubuntu_host_generation_output
+    BeakerHostGenerator::CLI.new(['ubuntu1404-64default.mdcal-ubuntu1404-64af']).execute!
+    assert_match(/WARNING: Starting with beaker-hostgenerator 1\.x platform strings for "el" hosts/, @stderr.string)
+  end
 end
--- a/lib/beaker-hostgenerator/cli.rb
+++ b/lib/beaker-hostgenerator/cli.rb
@@ -179,7 +179,7 @@ rather than "el-4-x86_64". It is recommended that you update your project's test
 suites ASAP or be forced to do so when beaker-hostgenerator development moves on
 to the 1.x series. We don't intend to backport features or platforms to 0.x.
 eow
-        STDERR.puts(warning)
+        $stderr.puts(warning)
       end
     end
 
--- a/spec/beaker-hostgenerator/generator_spec.rb
+++ b/spec/beaker-hostgenerator/generator_spec.rb
@@ -53,7 +53,7 @@ module BeakerHostGenerator
       arguments = fixture_hash["arguments_string"]
       it "beaker-hostgenerator #{arguments}" do
         arguments = arguments.split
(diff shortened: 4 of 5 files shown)


test/test_stdout.rb
```diff
@@ -18,4 +18,9 @@ def test_default_list_output
       assert_match(/.*#{spec_key}.*/, @stdout.string)
     end
   end
+
+  def test_ubuntu_host_generation_output
```
nwolfe: This doesn't really test the changes in this PR, it's just testing that this message is printed to stderr.

I think we want to test that stdout/stderr can be redirected in Ruby, which is what your PR is making possible.

So I'd think our test would want to override `$stdout`/`$stderr`, call `CLI.execute!` (with a much-simpler input; we don't care that it's ubuntu or anything about the input actually), then assert that the variables contain the message.

bin/genconfig2
```diff
@@ -5,7 +5,7 @@ $LOAD_PATH.unshift(lib_dir)
 
 require 'beaker-hostgenerator'
 
-STDERR.puts("Warning: 'genconfig2' is deprecated and will be removed in beaker-hostgenerator 1.x\n\n")
+$stderr.puts("Warning: 'genconfig2' is deprecated and will be removed in beaker-hostgenerator 1.x\n\n")
```
nwolfe: If we're going to change this for the stderr stream, we should definitely change it for all the streams (so also stdout).

So we should change `STDOUT` to `$stdout` everywhere too.

nwolfe:
FYI, this might be the time to just remove that warning, which we've already decided we're going to do in https://tickets.puppetlabs.com/browse/QENG-4019

These changes are still good though.

bin/genconfig2
ferglor: From what I can see, there are no direct references to STDOUT (or $stdout) in the lib, so I think we should be good here

test/test_stdout.rb
ferglor: So in this test I'm asserting the contents of @stderr, not $stderr, and in the setup phase of the test @stderr is initialised as `@stderr = StringIO.new`. I just chose the Ubuntu string because I knew it would hit the warning that gets written to stderr, should I just replace it with a simpler token? Something like `centos6-64ma`?

bin/genconfig2
nwolfe: Haha yeah my bad, I realized afterwards that we probably don't have any references to `$stdout` since they'd be hidden behind the calls to `puts`.

test/test_stdout.rb
nwolfe: Ah yeah didn't notice the `$` vs `@`, or the `setup` code.

I'm not sure that `setup` runs between each test though; I think it's just once at the beginning.
I would be worried about cross-test contamination.

Since our test is about redirecting the stream, I would expect it to explicitly:

1. Redirect `$stderr` to a new `StringIO`
2. Call `CLI#execute!`
3. Assert the stream contents

You might have to put the old stream back at the end of the test, or instead make `setup` run for every test rather than once.

[META: I'm probably going to be cleaning some of this up very soon when I get rid of the warning like I mentioned in that ticket, so don't worry too much about it being perfect.  I'm more interested in using this as an opportunity to share my thoughts on testing than I am about making this particular test better.]

nwolfe: > I just chose the Ubuntu string because I knew it would hit the warning that gets written to stderr, should I just replace it with a simpler token? Something like centos6-64ma?

The way I think about testing this is...

- The warning isn't platform-specific, so I would not want to make it seem like it is.

   So I would remove any mention of the platforms from the test name. 

- It also has nothing to do with the number of hosts or their roles, so I would choose the simplest input that would work.

   So I would simplify `'ubuntu1404-64default.mdcal-ubuntu1404-64af'` to `'ubuntu1404-64'`.

nwolfe:
See #70 -- we might not need this PR

ferglor:
@nwolfe I'm all for removing the warning message dude! Fewer lines ftw! But I'm guessing we're still going to have the genconfig2 warning in there?

nwolfe:
> But I'm guessing we're still going to have the genconfig2 warning in there?

Yeah but you shouldn't be calling `genconfig2` in Beaker for net-new code, right?

I'm good with changing that line from `STDERR` to `$stderr` though.  I wouldn't bother with a test if you make that change.

PR #70, opened 2017-02-01
(QENG-4019) Remove warning message about changing defaults

nwolfe:
This warning message was added a while ago to inform users of an upcoming
breaking change. We're no longer certain that we'll be making the breaking
change though, so the warning is a bit misleading.

Furthermore, we're now trying to silence this message when beaker-hostgenerator
is used programatically within Beaker.

There will likely be more changes that are outlined in the linked ticket, such
as adding a new CLI switch to toggle between specific & generic platform names.

### DIFF
--- a/lib/beaker-hostgenerator/cli.rb
+++ b/lib/beaker-hostgenerator/cli.rb
@@ -157,7 +157,6 @@ Usage: beaker-hostgenerator [options] <layout>
         templates = BeakerHostGenerator::AbsSupport.extract_templates(config)
         templates.to_json
       else
-        print_warnings
         config = BeakerHostGenerator::Generator.new.generate(@layout, @options)
         config.to_yaml
       end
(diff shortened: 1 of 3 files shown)


nwolfe:
@ferglor What if we just got rid of the warning message instead?  😄

nwolfe:
@waynr @johnduarte @jpartlow See https://tickets.puppetlabs.com/browse/QENG-4019 for the full story (see the comments).  Resurrecting an old topic here.

The last time we discussed switching the defaults, I believe we decided that it would actually hurt more than it would help.

The ticket discusses adding a CLI switch to give users the ability to choose whether they get generic or specific platform names -- that is not covered in this PR.

But we're now trying to work around this warning message in Beaker for the subcommand support:
- https://github.com/puppetlabs/beaker-hostgenerator/pull/69
- https://tickets.puppetlabs.com/browse/BKR-1032

I realized that we didn't need to do everything covered in the ticket at once, so rather than work around this warning message I figured it's time to just delete it.

waynr:
:+1:

PR #71, opened 2017-02-02
(BKR-1032) Switched from using STDERR to $stderr

ferglor:


ferglor:
Hilariously, this is the only change needed since QENG-4019 has went in 👍

nwolfe:
@ferglor I'm not against this change but I want to make sure you're not calling `genconfig2` in Beaker.  You shouldn't *need* to change this.

ferglor:
@nwolfe sorry dude, I forgot to comment on the other PR - you're right, I'm not calling `genconfig2`, I just thought I'd make this change since it's nicer to use $stderr over STDERR

PR #75, opened 2017-04-21
(maint) disks should be an array of integers

eputnam:
The ability to add disks to a vmpooler machine has been added to beaker and is set in the host file. Beaker is expecting an array of integers so if disks is present, this converts them to an array of integers before they're merged into the final config

### DIFF
--- a/lib/beaker-hostgenerator/generator.rb
+++ b/lib/beaker-hostgenerator/generator.rb
@@ -63,6 +63,7 @@ module BeakerHostGenerator
         # Merge in any arbitrary key-value host settings. Treat the 'hostname'
         # setting specially, and don't merge it in as an arbitrary setting.
         arbitrary_settings = node_info['host_settings']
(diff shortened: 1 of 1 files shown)


nwolfe:
@eputnam Nice.  I'd like to know more about this change though, is there more context in a JIRA ticket or something?  We just created a ticket to add some data structure support to BHG to allow for this disks support.

Also, unit tests!  We need to test all behavior.

eputnam:
indeed there is. in the acceptance tests for our red hat satellite module, we install satellite on one of the hosts. the default vmpooler box doesn't have enough available disk space so we just leveraged the existing vmpooler api to add disks. From there, we do all the disk mounting and expanding in the spec tests. In order to move the satellite module jenkins tests to use BHG, we need the ability to specify a disk in the host file, thus the need for this change.

Relevant JIRAZ: tickets.puppetlabs.com/browse/MODULES-4594, tickets.puppetlabs.com/browse/MODULES-4608
Relevant Gitz: https://github.com/puppetlabs/beaker/pull/1363

nwolfe:
Closing in favor of #78

PR #76, opened 2017-06-01
(QENG-4965) Add ABS support for hardware platforms

nwolfe:
TODO:
- [x] Consider adding an `:abs` key to every platform rather than (somewhat confusingly) reusing the `:vmpooler` data
- [x] Add unit test(s)
- [x] Update comments/docstrings/docs

----

```bash
$ beaker-hostgenerator aix53-POWERfa-aix72-POWERmdc-POWERa-POWERa-centos7-64a-64a --hypervisor abs --templates-only
{"aix-5.3-power":1,"aix-7.2-power":3,"centos-7-x86_64":2}
```


nwolfe:
@ferglor @mchllweeks Here's the BHG changes we need.  More work is necessary for it to be mergeable, but the changes we need are here so we can test out a more production-like scenario.

Since these changes aren't going out in a release yet, if we want to try them out in a CI pipeline, we'll need to probably make a new branch of whatever project we try running, and then in that branch make a commit that changes the Gemfile dependency to point to this branch of BHG so we can pick up the hardware support.

nwolfe:
@mchllweeks @ferglor This is ready for review / merge now.

We've been running with these changes since last week, and I just added more commits to update the tests, docstrings, and comments.  Also some minor cleanup commits.

nwolfe:
When this is merged I'll cut a release, and then all the projects *should* automatically pick up the latest version so we won't need to set the `BEAKER_HOSTGENERATOR_VERSION` ENV variable or point it to the github branch anymore.

PR #78, opened 2017-06-05
(QENG-4945) Add support for arbitrary lists

smcelmurry:
Before, we had no way to add lists to configurations. This adds a new bracket in order to parse out any lists the user may want to add in.

@nwolfe @eputnam

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,49 @@ CONFIG:
   pooling_api: http://vmpooler.delivery.puppetlabs.net/
 ```
 
+### Two hosts with arbitrary host settings with arbitrary lists
+
+```
+$ beaker-hostgenerator centos6-64m{disks=\[16\]-redhat7-64a{disks=\[8\,16\]}
+```
+
+Will generate
+
+```yaml
+---
+HOSTS:
+  centos6-64-1:
+    pe_dir:
+    pe_ver:
+    pe_upgrade_dir:
+    pe_upgrade_ver:
+    platform: el-6-x86_64
+    hypervisor: vmpooler
+    disks:
+    - 16
+    roles:
+    - agent
+    - master
+  redhat7-64-1:
+    pe_dir:
+    pe_ver:
+    pe_upgrade_dir:
+    pe_upgrade_ver:
+    hypervisor: vmpooler
+    platform: el-7-x86_64
+    template: redhat-7-x86_64
+    disks:
+    - 8
+    - 16
+    roles:
+    - agent
+CONFIG:
+  nfs_server: none
+  consoleport: 443
+  pooling_api: http://vmpooler.delivery.puppetlabs.net/
+
+
+
 ### Arbitrary global configuration settings
 
 ```
@@ -227,12 +270,14 @@ It may be necessary to URL-encode the input in order for it to properly be used
 in certain contexts, such as Jenkins.
 
 In most cases it will only be necessary to escape the characters that support
-arbitrary settings, which means the following four characters:
+arbitrary settings, which means the following characters:
 
 - `{` is `%7B`
 - `,` is `%2C`
 - `}` is `%7D`
 - ` ` is `%20`
+- `[` is `%5B`
+- `]` is `%5D`
 
 For a full URL encoding reference see: http://www.w3schools.com/tags/ref_urlencode.asp
 
--- a/lib/beaker-hostgenerator/cli.rb
+++ b/lib/beaker-hostgenerator/cli.rb
@@ -28,7 +28,7 @@ module BeakerHostGenerator
 Usage: beaker-hostgenerator [options] <layout>
 
  where <layout> takes the following form:
-  <platform>-<arch>[[<arbitrary-roles>,[...]].]<roles>[{<arbitrary-settings>,[...]}][[<arbirary-list>;[...]]][-<arch>[[<arbitrary-roles>,[...]].]<roles>[{<arbitrary-settings>,[...]}]][-<layout>]
+  <platform>-<arch>[[<arbitrary-roles>,[...]].]<roles>[{<arbitrary-settings>,[...]}][-<arch>[[<arbitrary-roles>,[...]].]<roles>[{<arbitrary-settings>,[...]}]][-<layout>]
 
  examples:
   centos6-64mdca-32a
@@ -54,13 +54,13 @@ Usage: beaker-hostgenerator [options] <layout>
    1 CentOS 6 64 bit node with roles = master, hypervisor = none, node name = static1, and my-key = my-value
    1 CentOS 6 32 bit node with roles = agent and the default hypervisor
 
-example with arbitrary host lists:
-  centos6-64m[disks=8.16;my-list=my-value1]-32a
-   1 CentOS 6 64 bit node with roles = master and lists:
+ example of a list within arbitrary host settings:
+  centos6-64m{hostname=static1\\,disks=[8,16],my-list=[my-value1]}-32a
+   1 CentOS 6 64 bit node with roles = master, node name = static1 and lists:
       disks:
          - 8
          - 16
-       my-key:
+       my-list:
          - my-value1
    1 CentOS 6 32 bit node with roles = agent and the default hypervisor
 
--- a/lib/beaker-hostgenerator/cli.rb
+++ b/lib/beaker-hostgenerator/cli.rb
@@ -28,7 +28,7 @@ module BeakerHostGenerator
 Usage: beaker-hostgenerator [options] <layout>
 
  where <layout> takes the following form:
-  <platform>-<arch>[[<arbitrary-roles>,[...]].]<roles>[{<arbitrary-settings>,[...]}][-<arch>[[<arbitrary-roles>,[...]].]<roles>[{<arbitrary-settings>,[...]}]][-<layout>]
+  <platform>-<arch>[[<arbitrary-roles>,[...]].]<roles>[{<arbitrary-settings>,[...]}][[<arbirary-list>;[...]]][-<arch>[[<arbitrary-roles>,[...]].]<roles>[{<arbitrary-settings>,[...]}]][-<layout>]
 
  examples:
   centos6-64mdca-32a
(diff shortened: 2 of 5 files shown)


spec/beaker-hostgenerator/parser_spec.rb
```diff
@@ -114,15 +114,16 @@ module BeakerHostGenerator
 
       context 'When using arbitrary host settings' do
         it 'Supports arbitrary whitespace in values' do
-          expect( parse_node_info_token("64{k1=value 1,k2=v2,k3=  v3  }") ).
+          expect( parse_node_info_token("64{k1=value 1,k2=v2,k3=  v3  ,k4=[v4~ v5 ~v6]}") ).
```
smcelmurry: Using the `~` here since we are calling the `parse_node_info_token` directly and the comma replacement does not happen beforehand.

README.md
```diff
@@ -157,6 +157,49 @@ CONFIG:
   pooling_api: http://vmpooler.delivery.puppetlabs.net/
 ```
 
+### Two hosts with arbitrary host settings with aribitrary lists
```
nwolfe: s/aribitrary/arbitrary

README.md
```diff
@@ -157,6 +157,49 @@ CONFIG:
...
+
+```
+$ beaker-hostgenerator centos6-64m{hypervisor=none\,hostname=static-master\,disks=\[16\]-redhat7-64a{disks=\[8\,16\]}
```
nwolfe: I'd simplify this so the new feature stands out more, so I'd probably remove the `hypervisor` and `hostname` settings from this

README.md
```diff
@@ -233,6 +276,9 @@ arbitrary settings, which means the following four characters:
 - `,` is `%2C`
 - `}` is `%7D`
 - ` ` is `%20`
+- `[` is `%5B`
+- `]` is `%5D`
+- `;` is `%3B`
```
nwolfe: We should delete the word "four" from line 273 so we don't have to update the count every time we update this list (and the count doesn't help anything anyway).

lib/beaker-hostgenerator/cli.rb
```diff
@@ -54,6 +54,16 @@ def initialize(argv = ARGV.dup)
    1 CentOS 6 64 bit node with roles = master, hypervisor = none, node name = static1, and my-key = my-value
    1 CentOS 6 32 bit node with roles = agent and the default hypervisor
 
+example with arbitrary host settings as a list:
```
nwolfe: This should be indented over a couple spaces so it's aligned with the other examples

nwolfe: Nitpicky but this kind of sounds like an alternate way to specify the arbitrary settings when we say "arbitrary host settings __as__ a list".

Maybe try rewording this so it's clear the word "list" is referring to the setting and not the outer map structure.

lib/beaker-hostgenerator/generator.rb
```diff
@@ -142,6 +142,10 @@ def unstringify_number_or_boolean(value)
         result = true
       elsif value == 'false'
         result = false
+      elsif value.kind_of?(Array)
+        value.each_with_index do |v, i|
+          result[i] = unstringify_number_or_boolean(v)
```
nwolfe: Nice!

We should update this method name though since it now supports more than just numbers and booleans.
Rather than spelling out the supported datatypes in the method name I think we can just call this `unstringify_value` which would go nicely with the function that calls it (`unstringify_values!`).

Also we should update the docstring on it.

lib/beaker-hostgenerator/parser.rb
```diff
@@ -110,14 +113,22 @@ def tokenize_layout(layout_spec)
...
           within_braces = false
+        when '['
+          within_brackets = true
+        when ']'
+          within_brackets = false
         when '-'
           spec[index] = '|' if within_braces
+        # replacing commas in arbirary list sections
+        when ','
+          spec[index] = '~' if within_brackets
```
nwolfe: Nicely done.

Should we update the docstring to account for this new stuff?

It does maybe kinda sorta feel like we're rewriting the code in the comment, but considering that the entire behavior is spread across several functions and it transitions through states it is nice to have a high-level description like this.

lib/beaker-hostgenerator/parser.rb
```diff
@@ -225,6 +235,11 @@ def settings_string_to_map(host_settings)
           raise BeakerHostGenerator::Exceptions::InvalidNodeSpecError,
                 "Malformed host settings: #{host_settings}"
         end
+        if pair[1].include?("[")
+          pair[1] = pair[1].
+          delete('[]').
+          split('~')
+        end
```
nwolfe: I'd pull this out of the validation loop and give it its own loop with a comment.

Also do we need to validate the elements of the list in any way?  Not sure, maybe not.

spec/beaker-hostgenerator/parser_spec.rb
nwolfe: I don't quite follow your comment.

How come we have to use the `~` directly; isn't it an implementation detail?

lib/beaker-hostgenerator/parser.rb
```diff
@@ -30,6 +30,9 @@ module Parser
     #   * host_settings    Any character (see `settings_string_to_map` for details)
     #                      Examples: {hostname=foo-bar, ip.address=123.4.5.6}
     #
+    #   * host_lists       Any character (see `settings_string_to_map` for details)
+    #                      Examples: [disks=16,8;other=foo,bar]
+    #
```
nwolfe: This is no longer its own thing, right?  List support is now just a value within the host_settings component right?

lib/beaker-hostgenerator/parser.rb
nwolfe: Are we sure that `~` is safe to use for this?  

First thing that comes to mind is filesystem paths; what if somebody wants to include an arbitrary setting like `foo_path: ~/foo` ?

lib/beaker-hostgenerator/parser.rb
smcelmurry: I don't think this needs extra validation as anything within the `[]` should technically work.

spec/beaker-hostgenerator/parser_spec.rb
smcelmurry: Because this test doesn't call `tokenize_layout`, this wouldn't be parsed properly and would split on the comma. I can add another spec test to make sure `tokenize_layout` is working as it should (i.e. producing the `~` on lists)

lib/beaker-hostgenerator/parser.rb
smcelmurry: It would only affect things inside of a list as the replace in `settings_string_to_map` doesn't look at anything outside the `[]`. But yeah for `home=[~/foo,bar]` is would break and split improperly. What symbol would suggest instead?

spec/beaker-hostgenerator/parser_spec.rb
```diff
@@ -113,16 +113,23 @@ module BeakerHostGenerator
...
             to eq({
                     "roles" => "",
                     "arbitrary_roles" => [],
                     "bits" => "64",
                     "host_settings" => {
                       "k1" => "value 1",
                       "k2" => "v2",
-                      "k3" => "  v3  "
+                      "k3" => "  v3  ",
+                      "k4" => ["v4"," v5 ","v6"]
```
nwolfe: I still don't understand why the `~` shows up in this file

lib/beaker-hostgenerator/parser.rb
```diff
@@ -100,16 +100,17 @@ def tokenize_layout(layout_spec)
...
       # And then finally turn the | back into - now that we've
       # properly decomposed the spec string:
-      #   ["centos6", "64{hostname=foo-bar}", "debian8", "32"]
+      #   ["centos6", "64m{hostname=foo-bar}", "debian8", "32"]
       #
       # NOTE we've specifically chosen to use the pipe character |
       # due to its unlikely occurrence in the user input string.
       spec = String.new(layout_spec) # Copy so we can replace characters inline
       within_braces = false
+      within_brackets = false
```
nwolfe: Leftovers

lib/beaker-hostgenerator/parser.rb
```diff
@@ -118,6 +119,7 @@ def tokenize_layout(layout_spec)
           within_braces = false
         when '-'
           spec[index] = '|' if within_braces
+        # replacing commas in arbirary list sections
```
nwolfe: Leftovers

lib/beaker-hostgenerator/parser.rb
```diff
@@ -206,18 +207,18 @@ def parse_node_info_token(token)
...
+    # @returns [Hash{String=>String|Array}] The host_settings string as a map.
     def settings_string_to_map(host_settings)
       # Strip it down to a list of pairs
       settings_pairs =
         host_settings.
         delete('{}').
-        split(',').
-        map { |keyvalue| keyvalue.split('=') }
-
+        split(/,(?=[^\]]*(?:\[|$))/)
```
nwolfe: Could we update the comment to try to explain what this regex is doing in English?

Also looks like we could re-attach the `.map` to this chain of calls again.

lib/beaker-hostgenerator/parser.rb
```diff
@@ -206,18 +207,18 @@ def parse_node_info_token(token)
...
+      settings_pairs = settings_pairs.map { |keyvalue| keyvalue.split('=') }
       # Validate they're actually pairs, and that all keys are non-empty
       settings_pairs.each do |pair|
         if pair.length != 2
+          puts pair.inspect
```
nwolfe: Leftovers

lib/beaker-hostgenerator/parser.rb
```diff
@@ -226,6 +227,14 @@ def settings_string_to_map(host_settings)
                 "Malformed host settings: #{host_settings}"
         end
       end
+      # Replace the remaining `,` for to make array for arbitrary list if brackets exist
```
nwolfe: s/for to/to

lib/beaker-hostgenerator/parser.rb
```diff
@@ -226,6 +227,14 @@ def settings_string_to_map(host_settings)
...
+      settings_pairs.each do |pair|
+        if pair[1].include?("[")
+          pair[1] = pair[1].
+          delete('[]').
+          split(',')
```
nwolfe: Since these two lines are a continuation they should be indented over so they're not at the same alignment as the top-level expression

spec/beaker-hostgenerator/parser_spec.rb
```diff
@@ -113,16 +113,23 @@ module BeakerHostGenerator
       end
 
       context 'When using arbitrary host settings' do
+        it 'Properly tokenize settings' do
```
nwolfe: We should be specific about supporting lists in this description rather than just referring them as general settings, so maybe throw in the word "list" or "array" somewhere in here?

lib/beaker-hostgenerator/generator.rb
```diff
@@ -118,30 +118,36 @@ def generate_host_roles!(host_config, node_info, options)
...
+        config['CONFIG'][k] = unstringify_value(v)
       end
     end
 
     # Attempts to convert numeric strings and boolean strings into proper
-    # integer and boolean types. Returns the input value if it's not a
-    # number string or boolean string.
+    # integer and boolean types. If value is an array, it will recurse
+    # through those values.
+    # Returns the input value if it's not a  number string or boolean string.
```
nwolfe: Extra space after "not a"

PR #88, opened 2017-08-14
(PE-21479) Return nil if version is an empty string

jpartlow:
Missed this case, which errors out because we end giving a nil
substitution to sprintf.

Also tighten up the case of an unparseable version string to return ''.

### DIFF
--- a/lib/beaker-hostgenerator/data.rb
+++ b/lib/beaker-hostgenerator/data.rb
@@ -25,7 +25,7 @@ module BeakerHostGenerator
     end
 
     def pe_dir(version)
-      return if version.nil?
+      return if version.nil? || version.empty?
 
       base_regex = '(\A\d+\.\d+)\.\d+'
       source = case version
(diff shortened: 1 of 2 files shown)


jpartlow:
Ok, this was actually a symptom of ci-job-configs not getting the  pe_upgrade_version from redis because of a flaw in puppetlabs/ci-job-configs#3174

But this should still be patched.  It's running here:

https://cinext-jenkinsmaster-enterprise-prod-1.delivery.puppetlabs.net/view/OPSWORK/job/enterprise_pe-acceptance-tests_integration-system_pe_opswork-upgrade_2017.2.x/20/LAYOUT=redhat6-64mcd%257Bamisize=c3.large%252Cvmname=amazon-6-x86_64%252Csnapshot=pe%252Cuser=ec2-user%257D,LEGACY_AGENT_VERSION=NONE,PLATFORM=NONE,SCM_BRANCH=2017.2.x,UPGRADE_FROM=NONE,UPGRADE_TO_VERSION=NONE,label=beaker-integration-smoke/console

PR #107, opened 2018-01-18
(maint) Correct the new fips platform to use el-7 as the platform

mchllweeks:
In order for Beaker to not have to know about all flavors of each OS, we should use el-7-x86_64 as the mapping for the platform and packaging_platform.

### DIFF
--- a/lib/beaker-hostgenerator/data.rb
+++ b/lib/beaker-hostgenerator/data.rb
@@ -625,7 +625,7 @@ module BeakerHostGenerator
         },
         'redhatfips7-64' => {
           :general => {
-            'platform'           => 'redhat_fips-7-x86_64',
+            'platform'           => 'el-7-x86_64',
             'packaging_platform' => 'redhat-fips-7-x86_64'
           },
           :vmpooler => {
--- a/test/fixtures/generated/default/redhatfips7-64aulcdfm
+++ b/test/fixtures/generated/default/redhatfips7-64aulcdfm
@@ -9,7 +9,7 @@ expected_hash:
       pe_upgrade_dir: 
       pe_upgrade_ver: 
       hypervisor: vmpooler
-      platform: redhat_fips-7-x86_64
+      platform: el-7-x86_64
       packaging_platform: redhat-fips-7-x86_64
       template: redhat-fips-7-x86_64
       roles:
--- a/test/fixtures/generated/multiplatform/redhatfips7-64aulcdfm-sles10-32-redhatfips7-64a
+++ b/test/fixtures/generated/multiplatform/redhatfips7-64aulcdfm-sles10-32-redhatfips7-64a
@@ -9,7 +9,7 @@ expected_hash:
       pe_upgrade_dir: 
       pe_upgrade_ver: 
       hypervisor: vmpooler
-      platform: redhat_fips-7-x86_64
+      platform: el-7-x86_64
       packaging_platform: redhat-fips-7-x86_64
       template: redhat-fips-7-x86_64
       roles:
@@ -36,7 +36,7 @@ expected_hash:
       pe_upgrade_dir: 
       pe_upgrade_ver: 
       hypervisor: vmpooler
-      platform: redhat_fips-7-x86_64
+      platform: el-7-x86_64
       packaging_platform: redhat-fips-7-x86_64
       template: redhat-fips-7-x86_64
       roles:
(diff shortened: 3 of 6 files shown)


ScottGarman:
Maybe I need a better understanding of how packaging_platform is used in beaker. I've assumed this will need to match the platform name that is used when building the agent in the puppet-agent repo. If that's not the case, this PR should work and won't require changes to beaker.

jsane:
Hmm... We do want to use FIPS enabled platform for packaging in puppet-agent. It is important since new vmpooler image (redhat-fips-7-x86_64) has a specific system openssl version that need to be aligned with what it would be at run-time. .i.e system openssl linked PA built on el-7 will not work on redhat-fips-7. 
I don't quite understand beaker enough to know how the proposed change will affect the above but it appears we will have issues.   /cc @underscorgan

mchllweeks:
@jsane I'm working with @kevpl to figure out the correct change here.  Will have an updated pr shortly.

joshcooper:
We use `packaging_platform` in beaker-puppet to know which platform the packaged artifact was built on, which may be different than the host we trying to install the artifact on. For example, on AIX 7.2, we install a package that was built on AIX 7.1. Similarly, on Windows we always build on 2012, but install on all windows variants. The way this fits together is:

During packaging, we generate a YAML file containing all of the build data. It contains an entry like. Note AIX doesn't support yum package manager (by default), so we don't generate repo_config info, just the built artifact.

```yaml
:platform_data:
  aix-6.1-power:
    :artifact: "./aix/6.1/PC1/ppc/puppet-agent-1.10.8.140.g1d90274-1.aix6.1.ppc.rpm"
    :repo_config:
```

During a CI run, we generate the beaker-hostgenerator data:

```
$ bundle exec beaker-hostgenerator aix72-POWERa
---
HOSTS:
  aix72-POWER-1:
    pe_dir:
    pe_ver:
    pe_upgrade_dir:
    pe_upgrade_ver:
    hypervisor: vmpooler
    platform: aix-7.2-power
    packaging_platform: aix-7.1-power
```

beaker-puppet uses the `packaging_platform` value, in this case `aix-7.1-power` and looks up the repo_config or artifact in the build data, which in this case is "./aix/6.1/PC1/ppc/puppet-agent-1.10.8.140.g1d90274-1.aix6.1.ppc.rpm". beaker-puppet then installs the artifact from that location on the build server.

On the other hand, beaker tests use `platform` to know what kind of platform the test is running on, and perform conditional logic. We can't rely on facter for that information, because we need to know what platform we're running on prior to installing puppet-agent (and facter).

mchllweeks:
@ScottGarman I updated the commit so that only the platform is changed to el-7.  This will allow beaker to treat it as a redhat platform, but still download the correct package.

ScottGarman:
Thanks for that great explanation @joshcooper. I'm :+1: on this PR.

jsane:
I understand the build platform can be different from the actual target platform as has been the case for some platforms. However since we are linking against system openssl, their versions happen to be different across el-7 (openssl-1.0.1e-60) and redhat-7 (openssl-1.0.1e-34) vmpooler images.  And that causes problems - it is not backwards ABI compatible when it should have been. Pl see https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/70 for the kind of failures. Here the agent, linked w/ system openssl, was built on el-7/centos-7 and targeted to run on redhat-7 w/o updating its system openssl.  
So as a workaround, till now, I used to bump the openssl version on redhat-7 to match that on centos-7 from within the various project's (puppet, pxp-agent, facter, cpp-pcp-client) setup/install scripts. And the hope/expectation was when we would build the agent on redhat-fips-7 we would not need that.

jsane:
Josh and me discussed this. Pl disregard my above comment. And take my comment with grain of salt. 
Since we are using the new fips platform for building we should be fine. While I don't completely understand the implications of using el-7 in the platform but may cause issues in some cases of testing using beaker for same reason of incompatible openssl versions (new fips platform has openssl verson 1.0.2k-8). We should be ok in CI where we get to specify the target platform - redhat-fips-7-x86_74. 
Although we would have liked to be able to take PA built on the redhat 7 fips platform and run it on other 'compatible' platforms (whether FIPS enabled or not) but unfortunately that does not possible.

PR #111, opened 2018-03-21
(IMAGES-793) Add the vro73-64 platform

ekinanp:


ekinanp:
This platform was added in exactly the same way as VRO 7 -- doing a `find . -name "*vro7*"` is where I got all the text fixtures from (along w/ a grep of vro7).

ekinanp:
Wondering if I should update the CHANGELOG.

ScottGarman:
Technically you should run the fixture generator to update the fixtures - there are instructions on how to do that in the BHG top-level README. This will re-generate all the fixtures, but you only want to add the ones relevant to your platform to this PR. Honestly I don't know that this difference really matters since the fixture generator creates random combinations of platforms. So I think this is good. :+1:

ekinanp:
So from @ScottGarman 's comments on https://github.com/puppetlabs/beaker-pe/pull/107, it seems like the only place I need to make a PR to enable beaker support for VRO 7.3 is here. Here's why:

1. https://github.com/puppetlabs/beaker/blob/master/lib/beaker/platform.rb#L6 does not contain VRO.

2. https://github.com/puppetlabs/beaker-pe/blob/master/lib/beaker-pe/pe-client-tools/install_helper.rb#L55 also has no checks for VRO.

3. https://github.com/puppetlabs/beaker-puppet/blob/master/lib/beaker-puppet/install_utils/foss_utils.rb#L255 also has no checks for VRO.

I'm going to double check with @jpadams on where they use beaker in the VRO pipelines, and also on my end to make sure VRO's platform name is not something else.

ekinanp:
I ran the same tests as https://cinext-jenkinsmaster-enterprise-prod-1.delivery.puppetlabs.net/view/vro/job/enterprise_puppet-vro-plugin_intn-sys_smoke-master/34/ locally and things work fine. So this is good to merge.

jpadams:
+1 LGTM

PR #116, opened 2018-04-20
(BKR-1435) Add the fedora28-64 platform

ekinanp:


ScottGarman:
Fedora 28 is only going to be a 64-bit platform, so you can remove the i386 versions.

ekinanp:
@ScottGarman Ok, I added the 32-bit one b/c the ticket mentioned both.

ekinanp:
Updated the PR to only add Fedora 28, 64.

ScottGarman:
Sorry for the confusion on the supported arches. Looks like the ticket generator was run with both arches listed, so all of the tickets are going to need to be updated.

PR #138, opened 2018-11-17
(BKR-1548) Totally disable getty when using docker

caseywilliams:
beaker-docker must run containers in privileged mode with an /sbin/init
entrypoint to successfully simulate a full operating sytem. This
combination can cause the container to open virutal consoles on the host
machine. These getty processes regularly consume 100% of CPU; they are
not automatically killed when the container is destroyed, and must be
cleaned up manually.

tty1 getty services were already disabled as part of the default set of
image commands for docker platforms, but this did not prevent the
behavior for other ttys. Replacing the getty executable with a copy of
/bin/true successfully disables all of them.

https://github.com/moby/moby/issues/4040 describes this issue and the workaround (and mentions beaker-docker explicitly).

### DIFF
--- a/lib/beaker-hostgenerator/data.rb
+++ b/lib/beaker-hostgenerator/data.rb
@@ -167,6 +167,7 @@ module BeakerHostGenerator
           },
           :docker => {
             'docker_image_commands' => [
+              'cp /bin/true /sbin/mingetty',
               'yum install -y crontabs initscripts iproute openssl sysvinit-tools tar wget which',
               'sed -i -e "/mingetty/d" /etc/inittab'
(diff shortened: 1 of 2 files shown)


smcelmurry:
The spec tests should be updated here to reflect your changes:
https://github.com/puppetlabs/beaker-hostgenerator/blob/2ffc6e815ae6e87208cd0b87bb771e868b1d7edd/test/fixtures/per-host-settings/every-hypervisor.yaml#L52-L54

caseywilliams:
Oops, thanks! I've updated the fixture and things are passing for me locally now

PR #139, opened 2018-11-20
Add a vagrant hypervisor

ekohl:
This also adds some base boxes based from the Vagrant Cloud.

Official builds are used for CentOS and Fedora in a generated way. For
Debian Wheezy, Jessie and Stretch 64 bits we also use the official
builds but manually since there's no automated way to map numbers to
releases. For all other platforms we rely on the generic builds which
uses the same naming scheme as our host specs.

We could use the official Ubuntu images in the same way as Debian, but
they're only built for Virtualbox where the generic ones are built for
other provides as well.

This PR is currently here for initial feedback and testing. I know I'll
need to add tests but an initial pass would be appreciated.

### DIFF
--- a/lib/beaker-hostgenerator/data.rb
+++ b/lib/beaker-hostgenerator/data.rb
@@ -297,6 +297,9 @@ module BeakerHostGenerator
               'apt-get update && apt-get install -y cron locales-all net-tools wget'
             ],
           },
+          :vagrant => {
+            'box' => 'debian/wheezy64',
+          },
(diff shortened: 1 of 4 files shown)


ekohl:
Updated. Now with vagrant_libvirt support and tests. Locally verified and should be ready to merge.

smcelmurry:
test this please

ekohl:
I now see I forgot a changelog entry.

PR #140, opened 2018-12-11
(QENG-7020) Add support for amazon linux 2 ARM64

smcelmurry:


test/fixtures/generated/osinfo-version-0/amazon7-ARM64m
```diff
@@ -0,0 +1,21 @@
+---
```
mattkirby: It looks like this and os-info-version-1/amazon7-arm64m are very similar. Can you explain why they're both needed?

smcelmurry: `osinfo-version-1` probably isn't needed as that is used for older versions of centos at this point. See here: https://github.com/puppetlabs/beaker-hostgenerator/blob/840545b098e0a3a29b79dfff68fc22ab0a5d003f/lib/beaker-hostgenerator/data.rb#L1580-L1639 Removing for clarity.

PR #144, opened 2018-12-18
(maint) Fix --hypervisor vagrant_libvirt

ekohl:
Previously the options[:hypervisor] did not end up in the hosts' config.  This refactors the hypervisors in a common method which always copies the original name into the resulting host config. It still prefers the explicit host hypervisor over the command line option.


ekohl:
Note I did take the liberty to refactor some internal code to reduce duplication. I did consider changing the `get_platform_info` to handle multiple hypervisors instead of doing that in the hypervisor class itself.

smcelmurry:
test this please

PR #145, opened 2019-01-30
(PA-2387) use el-7-aarch64 as platform for amazon7-ARM64

gimmyxd:
There is no platform(el-7-arm64) defined in puppet so we cannot build on that.
Build the packages for amazon7-ARM64 on el-7-aarch64 and test them on amazon7-ARM64.
I still need an opinion on this. Does it look ok?

adhoc build: https://jenkins-platform.delivery.puppetlabs.net/view/Adhoc/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/518/


gimmyxd:
can this be merged?

PR #146, opened 2019-02-18
(IMAGES-1028) Add beaker-hostgenerator support for Debian 10 Buster amd64

denis-susan:


gimmyxd:
You should generate the test fixtures using `rake generate:fixtures`, run the tests and commit only the new relevant fixtures.

PR #152, opened 2019-04-22
(QENG-6980) Add support for Cisco ios

mattkirby:
This commit adds support for Cisco ios and iosxe devices. Without this change beaker-hostgenerator does not support these devices.


mattkirby:
I'm not sure if this is correct for beaker-hostgenerator. These platforms will be served via nspooler.

PR #153, opened 2019-05-06
(QENG-7134) Individualize cisco IOS platforms

smcelmurry:
Before, these were all managed by one pool. These need to be called
individually based on their version number so there are now
separate calls for each version for cisco ios.

Additionally, made it clear that these are ARM32 hardware devices.


sbeaulie:
whoa

PR #171, opened 2019-10-14
(RE-12859) Migrate from enterprise.delivery.puppetlabs.net to Artifactory

mwaggett:
This commit replaces usage of enterprise.delivery.puppetlabs.net (aka saturn)
with Artifactory so that we can ultimately decommission it.

### DIFF
--- a/lib/beaker-hostgenerator/data.rb
+++ b/lib/beaker-hostgenerator/data.rb
@@ -14,7 +14,7 @@ module BeakerHostGenerator
   module Data
     module_function
 
-    PE_TARBALL_SERVER="http://enterprise.delivery.puppetlabs.net"
+    PE_TARBALL_SERVER="https://artifactory.delivery.puppetlabs.net/artifactory/generic_enterprise__local"
 
     def pe_version
       ENV['pe_version']
--- a/test/fixtures/generated/environment_variable_tests/pe_version_and_pe_family/centos6-32a
+++ b/test/fixtures/generated/environment_variable_tests/pe_version_and_pe_family/centos6-32a
@@ -8,9 +8,9 @@ environment_variables:
 expected_hash:
   HOSTS:
     centos6-32-1:
-      pe_dir: http://enterprise.delivery.puppetlabs.net/archives/releases/6.6.6/
+      pe_dir: https://artifactory.delivery.puppetlabs.net/artifactory/generic_enterprise__local/archives/releases/6.6.6/
       pe_ver: 6.6.6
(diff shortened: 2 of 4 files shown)


mwaggett:
I'm not totally sure how `pe_dir` is used, so I don't know if this will Just Work™ or not.

nmburgan:
I think this will work, but we'll need to get everything from 2016 forward into artifactory, or else those tests that use pre-Irving versions will fail.

mwaggett:
@nmburgan Do we need to be testing anything pre-Irving?? I know we have the one 2016.4 straggler, but besides that?

nmburgan:
Yeah, we have to test upgrades to the LTS, so we are upgrading from as old as 2016.4.15.

https://cinext-jenkinsmaster-enterprise-prod-1.delivery.puppetlabs.net/view/pe-integration/view/pe-2018.1.x/job/enterprise_pe-acceptance-tests_integration-system_pe_full-upgrade_nightly_2018.1.x/

mwaggett:
oh gross.

well those would be coming from `archives/releases` anyway, right? Those should already all be in Artifactory: https://artifactory.delivery.puppetlabs.net/artifactory/webapp/#/artifacts/browse/tree/General/generic_enterprise__local/archives/releases

nmburgan:
Oh right, good point :)

nmburgan:
I'm testing a fresh 2018.1.10 install and a 2016.4.15 -> 2018.1.10 now using this change.

nmburgan:
We should put a DNM (or I guess just not cut a gem?) until after the Lovejoy release.

mwaggett:
yeah, I have no idea how this gets released, but +1 to not shipping before Lovejoy, whether that means waiting to merge or waiting to release

nmburgan:
Verified that setting `pe_version` and `pe_upgrade_version` environment variables and then running `bundle exec beaker-hostgenerator centos7-64mcda --hypervisor=vmpooler` outputs a host file that points to artifactory for `pe_dir` and `pe_upgrade_dir`.  This is what CI does to generate the file.

Verified that that the install and upgrade noted above worked successfully pulling the tarballs from artifactory.

mwaggett:
@nmburgan @kevpl This should be good to go in, now that Lovejoy is done. No idea how this gets released though..

nmburgan:
Turns out we're going to need a quick 2019.2.1.  Let's wait until that goes out the door.

mwaggett:
okay how about now??

PR #177, opened 2020-01-15
Use debian/buster64 for Vagrant

ekohl:
This switches from generic/debian10 to debian/buster64. This is the official box and thus closer to what should be expected by end users.


sbeaulie:
ok to test

PR #180, opened 2020-03-05
install apt-transport-https on debian 8 and debian 9

Dan33l:
The package `apt-transport-https` does not look present on old docker images.

We have debian8 and debian9 jobs that are failing with message: `The method driver /usr/lib/apt/methods/https could not be found.`

Details :
https://github.com/voxpupuli/puppet-nodejs/pull/416

### DIFF
--- a/lib/beaker-hostgenerator/data.rb
+++ b/lib/beaker-hostgenerator/data.rb
@@ -514,7 +514,7 @@ module BeakerHostGenerator
             'docker_image_commands' => [
               'cp /bin/true /sbin/agetty',
               'rm -f /usr/sbin/policy-rc.d',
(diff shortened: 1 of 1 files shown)


ekohl:
This is likely due to a regression in puppetlabs-apt 7.4.0 (https://github.com/puppetlabs/puppetlabs-apt/commit/f862889455c4c9994d0afc21a84715f4bc051a9c) when lsb-release is not installed. In that case the `lsbdistrelease` fact is missing and puppetlabs-apt no longer installs the dependency.

IMHO the images should only install the bare minimum

ekohl:
It was pointed out to me that the puppet-release deb has switched its URLs to https so Puppet itself wasn't getting installed. In that case, :+1: since it fixes a bootstrap issue.

PR #182, opened 2020-04-07
Add support for latest OpenSuSE images

dhoppe:
I would like to use the latest OpenSuSE images to test different Puppet modules using Travis CI. 

At least ```Docker``` and ```Vagrant``` should work without any problemt, but I am not sure if additional ```docker_image_commands``` are necessary.


ekohl:
Could you rebase this?

PR #194, opened 2020-10-19
Generate Fedora and CentOS data where possible

ekohl:
All Fedora definitions are the same, so they might as well be generated in a loop. packaging_platform is dropped since beaker defaults to platform. In the case of Fedora they always match.

For CentOS only the vmpooler templates are generated because the Docker image commands differ too much.


genebean (approved):

genebean:
Merging but not pushing a new gem yet due to so many people on PTO for Thanksgiving and such.

ekohl:
No problem, glad it's at least merged now.

PR #198, opened 2020-11-25
Add Fedora 33 support

ekohl:
Even though no offical packages have been built by Puppet, it still makes sense to be able to generate a definition.


genebean (approved):

genebean:
Merging but not releasing a gem right now due to soooo many people being on holiday.

PR #203, opened 2021-02-10
(BKR-1698) Use generic vagrant box for centos6

wbclark:
The centos organization on vagrant cloud (htttps://app.vagrantup.com/centos)
no longer provides a centos/6 box. This commit switches to the generic base
boxes for centos which are still available and kept up to date.

### DIFF
--- a/lib/beaker-hostgenerator/hypervisor/vagrant.rb
+++ b/lib/beaker-hostgenerator/hypervisor/vagrant.rb
@@ -8,7 +8,7 @@ module BeakerHostGenerator
       include BeakerHostGenerator::Data
 
       def generate_node(node_info, base_config, bhg_version)
-        if node_info['ostype'] =~ /^centos/
+        if node_info['ostype'] =~ /^centos/ and not node_info['ostype'] =~ /6/
           base_config['box'] = node_info['ostype'].sub(/(\d)/, '/\1')
         elsif node_info['ostype'] =~ /^fedora/
           base_config['box'] = node_info['ostype'].sub(/(\d)/, '/\1') + '-cloud-base'
--- a/test/fixtures/per-host-settings/vagrant-hypervisor.yaml
+++ b/test/fixtures/per-host-settings/vagrant-hypervisor.yaml
@@ -8,7 +8,7 @@ expected_hash:
       pe_ver:
       pe_upgrade_dir:
       pe_upgrade_ver:
-      box: centos/6
+      box: generic/centos6
       synced_folder: disabled
       platform: el-6-x86_64
       packaging_platform: el-6-x86_64


wbclark:
I've updated this to preserve the current behavior for centos7 and centos8, and use the generic box for centos6 only

ekohl (changes requested):
I don't think this is wise. CentOS 6 is dead. Do not attempt to keep it alive IMHO. It's not like you still have a mirror network so you can't install anything unless you change it to the non-mirrored vault. The OS is EOL and deserves to die. It's not like we made any effort to keep CentOS 5 alive either.

wbclark:
> I don't think this is wise. CentOS 6 is dead. Do not attempt to keep it alive IMHO. It's not like you still have a mirror network so you can't install anything unless you change it to the non-mirrored vault. The OS is EOL and deserves to die. It's not like we made any effort to keep CentOS 5 alive either.

Hey @ekohl , thanks for the review.

I'm in 100% agreement with you when it comes to dropping EL6 support in `theforeman/puppet-*`. We shouldn't for any reason continue to support it. I would be happy to move forward with that today since it seems we are all onboard.

I am a lot less confident in saying that we should prevent beaker-hostgenerator users from testing against EL6 under any circumstances, given that:

1. The Centos6 box not only exists at https://app.vagrantup.com/generic/boxes/centos6 but is also still receiving updates (v3.2.6 released 12 days ago, another as yet unreleased version 3.2.8 created 6 hours ago at the time I am writing this)

2. We can't know the circumstances of every user that might need to develop some puppet modules to manage EL6 systems that they need to maintain for a while longer because of whatever circumstance (that is likely beyond their individual control).

I'm not sure if voxpupuli has any guideline or best practice in this area, and if so then of course I'll defer to it. In the absence of an established rule for dealing with this sort of scenario, my feeling is that the image exists and we shouldn't try to decide for users that they are not allowed to test with it, when the user is much closer to their 'need' or 'business case' for doing so than we are.

ekohl:
>     1. The Centos6 box not only exists at https://app.vagrantup.com/generic/boxes/centos6 but is also still receiving updates (v3.2.6 released 12 days ago, another as yet unreleased version 3.2.8 created 6 hours ago at the time I am writing this)

These are non-official boxes. They do have changes from official images (in my experience most notably with using Google DNS by default which can break in corporate firewall setups).

I also don't see any changes in https://github.com/lavabit/robox/blob/master/http/generic.centos6.vagrant.ks to use vault as repositories.

>     2\. We can't know the circumstances of every user that might need to develop some puppet modules to manage EL6 systems that they need to maintain for a while longer because of whatever circumstance (that is likely beyond their individual control).

IMHO the burden is on these users. They can also switch to `centos6-64{image=generic/centos6}` as a workaround.

> I'm not sure if voxpupuli has any guideline or best practice in this area, and if so then of course I'll defer to it. In the absence of an established rule for dealing with this sort of scenario, my feeling is that the image exists and we shouldn't try to decide for users that they are not allowed to test with it, when the user is much closer to their 'need' or 'business case' for doing so than we are.

There are strong proponents of killing anything that's EOL. Generally the project has a hard enough time to support what is still upstream supported and there's little to no capacity for anything EOL. While it hasn't been done here, it may actually mean removing older operating systemes that are no longer supported.

Other workarounds include storing the YAML setfile in the repository with modifications. Remember that beaker-hostgenerator is essentially a templating engine which feeds input to beaker itself.

All in all I think we shouldn't put in any effort to support CentOS 6 when upstream has made it very clear that it should no longer be used by removing it from their mirror network.

wbclark:
OK. Thanks for the explanation @ekohl and I'm satisfied by that reasoning.

> While it hasn't been done here, it may actually mean removing older operating systemes that are no longer supported.

In that case should we simply close this PR, or is it preferable to rework it to remove any references to EL6?

ekohl:
I think we should close it. We didn't remove 5 either. If we do decide to clean up, it should be The Great Cleanup IMHO.

PR #206, opened 2021-04-12
Modify osx11.2 to osx11

BobosilaVictor:
Related to : https://github.com/puppetlabs/puppet-agent/pull/2081

### DIFF
--- a/lib/beaker-hostgenerator/data.rb
+++ b/lib/beaker-hostgenerator/data.rb
@@ -803,10 +803,10 @@ module BeakerHostGenerator
             'template' => 'osx-1015-x86_64'
           }
         },
-        'osx112-64' => {
+        'osx11-64' => {
           :general => {
-            'platform'           => 'osx-11.2-x86_64',
-            'packaging_platform' => 'osx-11.2-x86_64'
+            'platform'           => 'osx-11-x86_64',
+            'packaging_platform' => 'osx-11-x86_64'
           },
           :vmpooler => {
             'template' => 'macos-112-x86_64'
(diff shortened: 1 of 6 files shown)


GabrielNagy (approved):
For Big Sur we're making the assumption (for now) that all Y releases of macOS 11 will still refer to Big Sur (as has been the case with 11.1, 11.2, and 11.3 beta). This is to avoid having to add a new platform for each minor release of Big Sur.

The vmpooler template should remain the same, as it reflects the underlying version of the machine we're building on, but the platform should be agnostic.

ekohl (commented):
I'm not familiar with OS X but does it make sense to keep the old definition?

GabrielNagy:
I don't think so, right now keeping 11.2 would be like having separate definitions for RedHat minor releases (like 8.1, 8.2, 8.3).

However... this is definitely something that might change at any point, especially since Apple doesn't communicate versioning details to external users. We can circle back if/when something changes.

ekohl:
If supported minors aren't a thing for OS X I don't mind removing it. However, it is odd that for OS X 10 there are multiple minor versions. Is that a new policy for 11?

GabrielNagy:
Ah, yes, they've decided to bump the major version for the first time in ~20 years 😄 

- https://en.wikipedia.org/wiki/MacOS_version_history#Version_11:_%22Big_Sur%22
- https://www.macworld.com/article/342024/after-20-years-of-os-x-macos-11-might-be-gone-after-big-sur.html

ekohl (approved):

ciprianbadescu:
@ekohl, do you know when could we get a release with MacOS 11 support?
Thanks

bastelfreak:
I raised https://github.com/voxpupuli/beaker-hostgenerator/pull/207 for a release

ciprianbadescu:
@bastelfreak, much obliged

PR #210, opened 2021-05-10
(IMAGES-1273) add support for Debian 11 amd64

ciprianbadescu:


ekohl (changes requested):
There is already a vagrant box (https://app.vagrantup.com/debian/boxes/bullseye64). Please add that as well (like we did with Debian 10). It would also be nice if we could get the correct Docker image commands but I won't block on that.

bastelfreak (approved):

ekohl (approved):
Thanks!

PR #211, opened 2021-05-11
release 1.4.0

bastelfreak:


bastelfreak:
@ekohl I'm not sure if we should mark `Deal with nil when merging` as bug or if it's fine  as it is.

PR #209, opened 2021-04-28
Deal with nil when merging

ekohl:
This does at least make rake generate:fixtures work, but then the result is still not really what you'd expect.

PR #213, opened 2021-05-13
Avoid activesupport deep_merge conflict

GabrielNagy:
github-changelog-generator depends on activesupport which defines `Hash#deep_merge` methods that behave differently and conflict with the ones from the `deep_merge` gem. Use the recommended way of avoiding this conflict, as per the `deep_merge` documentation.[1]

Also, revert the changes from #209 which cause unexpected behavior with generated fixtures.

[1] https://github.com/danielsdeleo/deep_merge#using-deep_merge-in-rails

### DIFF
--- a/lib/beaker-hostgenerator/generator.rb
+++ b/lib/beaker-hostgenerator/generator.rb
@@ -22,7 +22,7 @@ module BeakerHostGenerator
     def generate(layout, options)
       layout = prepare(layout)
       tokens = tokenize_layout(layout)
-      config = {}.deep_merge(BASE_CONFIG)
+      config = {}.deeper_merge(BASE_CONFIG)
       nodeid = Hash.new(1)
       ostype = nil
       bhg_version = options[:osinfo_version] || 0
@@ -51,7 +51,7 @@ module BeakerHostGenerator
         # Delegate to the hypervisor
         hypervisor = BeakerHostGenerator::Hypervisor.create(node_info, options)
         host_config = hypervisor.generate_node(node_info, host_config, bhg_version)
-        config['CONFIG'].deep_merge!(hypervisor.global_config())
+        config['CONFIG'].deeper_merge!(hypervisor.global_config())
 
         # Merge in any arbitrary key-value host settings. Treat the 'hostname'
         # setting specially, and don't merge it in as an arbitrary setting.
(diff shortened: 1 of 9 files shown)


lib/beaker-hostgenerator/generator.rb
```diff
@@ -22,7 +22,7 @@ class Generator
     def generate(layout, options)
       layout = prepare(layout)
       tokens = tokenize_layout(layout)
-      config = {}.deep_merge(BASE_CONFIG)
+      config = {}.deeper_merge(BASE_CONFIG)
```
ekohl, line 25: Should this be just
```suggestion
      config = BASE_CONFIG.dup
```

Or does `deeper_merge` also duplicate `CONFIG` within `BASE_CONFIG`?

Taking a deeper look: what is `consoleport`? In none of the beaker projects I can find any documentation. Is it something that only one provider (abs?) needs? Safe for nfs_server.

ekohl (approved):

lib/beaker-hostgenerator/generator.rb
GabrielNagy: Honestly I have no idea... `consoleport` sounds like it could be a PE-specific thing, but I 
don't know about `nfs_server`. I don't see them used anywhere in our projects... maybe it would make sense to remove both in a major release.

> Or does `deeper_merge` also duplicate `CONFIG` within `BASE_CONFIG`?

That seems to be the case, object IDs are different with `deeper_merge`:
```
[20] pry(#<BeakerHostGenerator::Generator>)> BASE_CONFIG.dup["CONFIG"].object_id == BASE_CONFIG["CONFIG"].object_id
=> true
[21] pry(#<BeakerHostGenerator::Generator>)> {}.deeper_merge(BASE_CONFIG)["CONFIG"].object_id == BASE_CONFIG["CONFIG"].object_id
=> false
```

ekohl: Looks like consoleport was introduced [here](https://github.com/voxpupuli/beaker/commit/7492b5b6415fff2e4d525a3ccb5b76537d45179d) and nfs_server [here](https://github.com/voxpupuli/beaker/commit/f6523410f3bb57e4421c9856f96d6798dbc96adf). Both are PE-specific. I think we should do a major version where we drop them. It's adding a lot of noise for something that's barely if at all used.

PR #218, opened 2021-07-28
Generate Ubuntu ABS, vmpooler, and Docker templates on the fly

ekohl:
This is similar to 744532d309d370fdea636f2b14b8e7ed2ccaf882 but less complete.

Currently some tests are still failing.


GabrielNagy:
The remaining tests fail because Ubuntu platforms with exotic architectures (other than 32/64 bit) did not have a `template` key for vmpooler before. I think it's fine to update the fixtures to include the template key.

ekohl:
If you have the time, please do. It'd be great if you could even generate some more in a loop, but I'm not sure how well that would work given there are some docker commands needed.

GabrielNagy:
I think this should be good to go, I did the cleanup/generation/fixture adding in separate commits and I checked that specs passed after each run. Apart from changes in the docker package install order things seem to be the same.

GabrielNagy:
This PR would also render https://github.com/voxpupuli/beaker-hostgenerator/pull/217 useless 😄

lib/beaker-hostgenerator/hypervisor/docker.rb
```diff
@@ -17,6 +17,33 @@ def generate_node(node_info, base_config, bhg_version)
...
+          extra_packages_to_install = case node_info['ostype']
+                                      when /1404/
+                                        []
+                                      when /1604/
+                                        ['locales']
+                                      else
+                                        ['locales', 'iproute2', 'gnupg']
+                                      end
+
+          docker_commands << "apt-get install -y net-tools wget apt-transport-https #{extra_packages_to_install.join(' ')}"
```
ekohl, line 40: apt-transport-https a transitional package starting at least 16.04: https://packages.ubuntu.com/bionic/apt-transport-https. It was still a real package in Debian 9 but in Debian 10 it's also a transitional package.

lib/beaker-hostgenerator/hypervisor/docker.rb
```diff
@@ -17,6 +17,33 @@ def generate_node(node_info, base_config, bhg_version)
           base_config['image'] = "amd64/#{base_config['image']}"
         end
 
+        docker_commands = []
+
+        case node_info['ostype']
+        when /^ubuntu/
+          docker_commands << 'cp /bin/true /sbin/agetty'
+
+          if node_info['ostype'] =~ /14.04/
```
ekohl, line 26: Here you use `14.04` but I think this should be 1404.

ekohl (commented):
Thanks!

lib/beaker-hostgenerator/hypervisor/docker.rb
GabrielNagy: yeah it seems to be a noop on 16.04 and newer, good catch!

ekohl (commented):
I can't approve my own PR, but :+1:

GabrielNagy (approved):

PR #226, opened 2021-09-20
(IMAGES-1302) Add Alma and Rocky 8 support

GabrielNagy:
Add Alma and Rocky support, and update the vmpooler generator to be able to handle these OSes.

Update all modified fixtures.


lib/beaker-hostgenerator/hypervisor/vmpooler.rb
```diff
@@ -18,10 +18,10 @@ def generate_node(node_info, base_config, bhg_version)
         base_config = base_generate_node(node_info, base_config, bhg_version, :vmpooler)
 
         case node_info['ostype']
-        when /^centos/
-          base_config['template'] = base_config['platform'].gsub(/^el/, 'centos')
+        when /^(almalinux|centos|redhat|rocky)/
+          base_config['template'] ||= base_config['platform'].gsub(/^el/, $1)
```
GabrielNagy, line 22: I've updated this logic to only modify the `template` key in case it's not defined in the base config. This seems like a sensible approach to me, allowing overrides on a case-by-case basis.

bastelfreak (approved):

PR #236, opened 2021-12-05
Drop legacy genconfig2 cli tool

bastelfreak:
The tool is deprecated, it also had a deprecation notice for a long
time.


bastelfreak:
I created this PR for visibility. I would like to merge when a major release is required and *not now*

PR #239, opened 2022-01-27
Drop redundant attributes from generated data

ekohl:
This drop packaging_platform where it matches platform. Beaker already returns the platform when packaging_platform isn't set. This means specifying if it matches is redundant.

It also omits the empty entries for pe_*.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,7 @@ source ENV['GEM_SOURCE'] || 'https://rubygems.org'
 
 gemspec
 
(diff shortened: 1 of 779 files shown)


bastelfreak (approved):
Looks fine to me. Maybe @GabrielNagy wants to take a look at well

GabrielNagy (approved):
LGTM!

We changed the logic in our internal CI to fall back to `platform` if `packaging_platform` doesn't exist, so this should (hopefully) not have any negative effects.

PR #242, opened 2022-02-01
Install hostname on CentOS 8 Stream in docker

ekohl:
This is not available in the minimal CentOS 8 Stream containers and specinfra relies on this for the hostname -f command to determine the FQDN.

Fixes: a4bdd3b3c51ccc65d6e75a9fa245a2f240b0d12d


evgeni:
I have no powers in this world, but the change looks correct to me :)

bastelfreak (approved):

PR #244, opened 2022-02-14
Install lsb-release on Ubuntu 16/18 docker images

root-expert:
`lsb_release` tools is now missing from Ubuntu 16/18 docker images causing acceptance puppet 6 tests to fail.
Facter 3.x utilizes `lsb_release` tool for obtaining `os.distro` fact.

Some PRs that are failing due to this:

* https://github.com/voxpupuli/puppet-zabbix/pull/813
* https://github.com/voxpupuli/puppet-cvmfs/pull/148 (workaround on this specific module)
* https://github.com/voxpupuli/puppet-nginx/pull/1496

And probably all of the modules that use `os.distro` fact will fail on Ubuntu 16/18 platforms.

P.S it's the first time touching this utility so any help is appreciated it :smile:

### DIFF
--- a/lib/beaker-hostgenerator/hypervisor/docker.rb
+++ b/lib/beaker-hostgenerator/hypervisor/docker.rb
@@ -32,9 +32,9 @@ module BeakerHostGenerator
                                       when /1404/
                                         ['apt-transport-https']
                                       when /1604/
-                                        ['locales']
+                                        ['locales', 'lsb-release']
                                       else
-                                        ['locales', 'iproute2', 'gnupg']
+                                        ['locales', 'iproute2', 'gnupg', 'lsb-release']
                                       end
 
           docker_commands << "apt-get install -y net-tools wget #{extra_packages_to_install.join(' ')}"
--- a/test/fixtures/per-host-settings/docker-regex-validation.yaml
+++ b/test/fixtures/per-host-settings/docker-regex-validation.yaml
@@ -28,7 +28,7 @@ expected_hash:
       platform: ubuntu-20.04-amd64
       docker_image_commands:
       - cp /bin/true /sbin/agetty
-      - apt-get install -y net-tools wget locales iproute2 gnupg
+      - apt-get install -y net-tools wget locales iproute2 gnupg lsb-release
       - locale-gen en_US.UTF-8
       - echo LANG=en_US.UTF-8 > /etc/default/locale
       hypervisor: docker


ekohl (changes requested):
I wonder if this really works. It's not a deep merge so this part will overwrite it:
https://github.com/voxpupuli/beaker-hostgenerator/blob/08f607482950966e80b76eedbe303c7d0351502f/lib/beaker-hostgenerator/hypervisor/docker.rb#L40-L42

Please change it there.

root-expert:
Wouldn't the rest `docker_image_commands` get overridden by that as well? E.g 

https://github.com/voxpupuli/beaker-hostgenerator/blob/08f607482950966e80b76eedbe303c7d0351502f/lib/beaker-hostgenerator/data.rb#L660-#L664

ekohl (commented):
Taking a step back: doing so gives the impression you can always count on these facts. That can lead to subtle breakages. For example, I think in the case of Zabbix the solution is to use safer methods that don't fail as hard (https://github.com/voxpupuli/puppet-zabbix/pull/813#issuecomment-1039308906). Nginx now passes so I no longer see why it failed. It may need a similar treatment.

I think I'm generally more in favor of that than pretending it's always present. For example, now it won't be in the README and users may run into it and find out the hard way.

root-expert:
IMO these facts on Puppet 7 are supposed to be always present like `os.family` because their value is based on `/etc/os-release` unlike with Puppet 6 which is based on `lsb_release`.

I agree, though, that accessing facts in a more secure way should become a coding guideline here in Voxpupuli. E.g we already have the "no topscope variables for facts" maybe we should enforce `fact()` or `dig`, as you said, as well.

I still believe this PR still has its value since multiple modules' PRs are failing due to this right now and migrating away from these facts or accessing them with `fact()` is time consuming.

root-expert:
That's a good point actually. I believe that `lsb-release` should be a recommended package for `puppet-agent` since `os.distro` facts are not legacy and Facter 3 needs lsb-release to resolve them. But I don't think that's gonna happen.

Linting this is a good idea for Voxpupuli. Puppet Inc. uses these facts on some their modules as well:

* [puppetlabs-apt](https://github.com/puppetlabs/puppetlabs-apt/blob/main/manifests/source.pp#L82-L83)
* [puppetlabs-postgresql](https://github.com/puppetlabs/puppetlabs-postgresql/blob/main/manifests/repo/apt_postgresql_org.pp#L19)

So some our workflows will still fail because of external modules.

baurmatt:
Soooooo, what's the recommended solution for this? :) puppet-prometheus is running into this as well.

For me it sounds totally valid that hostgenerator adds lsb-release for Ubuntu 16.04/18.04 as those Docker images contained it before.

ekohl:
I'm still :-1: on this. If a module needs lsb-release, it should clearly state this and deal with it. It's not something you can count on. I don't have `lsb-release` installed on my actual systems (VMs) either.

PR #249, opened 2022-05-20
(IMAGES-1339) Add Ubuntu 22.04 support

cthorn42:
The Ubuntu 22.04 VMs are hosted on GCE, which has a different template.
Because the way the template is defined for ubuntu we have to add some logic
that if the Ubuntu host is 2204, then it has a slightly different template.


cthorn42:
Tested this locally:
{code}
./bin/beaker-hostgenerator ubuntu2204-64
---
HOSTS:
  ubuntu2204-64-1:
    platform: ubuntu-22.04-amd64
    hypervisor: vmpooler
    template: ubuntu-2204-x86_64-x64-gce-n1-standard-2
    roles:
    - agent
CONFIG:
  nfs_server: none
  consoleport: 443
  pooling_api: https://vmpooler-prod.k8s.infracore.puppet.net/
{code}
And for non 2204:
{code}
./bin/beaker-hostgenerator ubuntu1604-64
---
HOSTS:
  ubuntu1604-64-1:
    platform: ubuntu-16.04-amd64
    hypervisor: vmpooler
    template: ubuntu-1604-x86_64
    roles:
    - agent
CONFIG:
  nfs_server: none
  consoleport: 443
  pooling_api: https://vmpooler-prod.k8s.infracore.puppet.net/
{code}

cthorn42:
Doh, had an error in the template, pushed up a commit so now it correctly reads as:
```
---
HOSTS:
  ubuntu2204-64-1:
    platform: ubuntu-22.04-amd64
    hypervisor: vmpooler
    template: ubuntu-2204-x86_64-gce-n1-standard-2
    roles:
    - agent
CONFIG:
  nfs_server: none
  consoleport: 443
  pooling_api: https://vmpooler-prod.k8s.infracore.puppet.net/
```

joshcooper:
It's been awhile since I've worked with this repo... Is the same change needed in the `abs` hypervisor too https://github.com/voxpupuli/beaker-hostgenerator/blob/afd6f05c253d1422793249192221064143ae16d1/lib/beaker-hostgenerator/hypervisor/abs.rb#L45

joshcooper:
Cool, cool. The change makes sense, but on second thought, I'm wondering if abs & vmpooler should expose this detail to clients? In other words, when I request `redhat-9-x86_64`, it's actually an alias for the `redhat-9-x86_64-pixa4` template:

https://github.com/puppetlabs/dio-helm-charts/blob/d3fdf047c3724093b118bbc63abd3a7964ff8d67/charts/dio-vmpooler/templates/configmap-dio-vmpooler-pools.yaml#L568-L569

I'm wondering if we should do the same here? Basically add `ubuntu-2204-x86_64` as an alias for the `ubuntu-2204-x64-gce-n1-standard-2` host, so that clients like `beaker-hostgenerator` don't need to modify the template name?

cthorn42:
Good question, if we wanted to just use `ubuntu-2204-x86_64` as the template name, I say we should just change https://github.com/puppetlabs/dio-helm-charts/blob/master/charts/dio-vmpooler/templates/configmap-dio-vmpooler-pools.yaml#L183 to that instead of worrying about the alias.
Of course there was some concern that right now we have not officially "cut over" to GCE images. This VM will be kind of a back dooring us until the process. I was being cautious by using the long GCE name to be explicit that this was a GCE image.  
I'm open to going with the simplier name, but I guess we'd have to send out an e-mail notifying people that it is a GCE vmpooler VM and not a standard locally hosted vmpooler VM.

joshcooper:
> but I guess we'd have to send out an e-mail notifying people that it is a GCE vmpooler VM and not a standard locally hosted vmpooler VM.

That's partly why I think having the alias and non-alias names are useful. If you don't care where the VM is hosted, you can request the short name `ubuntu-1804-x86_64`. If you really care about where the VM is hosted, you can request it explicitly `ubuntu-1804-x86_64-pixa4` or `ubuntu-1804-x64-gce-n1-standard-2`.

Currently `ubuntu-1804-x86_64` is an alias for the former, but as we migrate things over, we can change the alias in one place (in the vmpooler config) to point the alias to the latter.

It also looks like the GCE VMs are using a different architecture naming scheme (`x86_64` vs `x64`) do you know if that's intentional? For example:

```
ubuntu-1804-x64-gce-n1-standard-2
ubuntu-1804-x86_64-pixa4
```

joshcooper:
> A shorter alias is IMHO out of scope for beaker-hostgenerator. Otherwise the names proliferate too much.

@ekohl I didn’t quite follow this part? I was suggesting we update beaker-hostgenerator to understand the short name, eg it can translate ubuntu2204-64 to the template ubuntu-2204-x86_64, but if someone wanted to override the template, then they could do what you suggested. So I think we’re saying the same thing?

cthorn42:
In the case of `ubuntu-2204-x86_64` there is only one option for our use case, GCE. So I'm less inclined to go the alias route.
As for architecture naming scheme, I think that is something the DIO team did on their own without the context of beaker-hostgenerator template naming. I think it is slightly off and we should correct them on it going forward.

ekohl:
> > A shorter alias is IMHO out of scope for beaker-hostgenerator. Otherwise the names proliferate too much.
> 
> @ekohl I didn’t quite follow this part? I was suggesting we update beaker-hostgenerator to understand the short name, eg it can translate ubuntu2204-64 to the template ubuntu-2204-x86_64, but if someone wanted to override the template, then they could do what you suggested. So I think we’re saying the same thing?

It sounds like we are indeed saying the same thing.

mhashizume (approved):

PR #255, opened 2022-06-27
(IMAGES-1297) Adds macOS 11/12 M1 support

mhashizume:
For both IMAGES-1297 and IMAGES-1378, adds support for macOS 11
(Big Sur) and 12 (Monterey) M1 (ARM64) platforms.


joshcooper (approved):
LGTM I was surprised by the uppercase `-ARM64` but I see BHG uses that for amazon linux and others like POWER, etc, so :+1:

PR #265, opened 2022-12-14
AlmaLinux: Fix docker image names

bastelfreak:


ekohl (commented):
Shouldn't this (and Rocky) be fixed here:
https://github.com/voxpupuli/beaker-hostgenerator/blob/d0372c6b7e76f0797bc0639de9d3346ea5dd00de/lib/beaker-hostgenerator/hypervisor/docker.rb#L12-L15

By now you could probably make it a case statement.

ekohl (approved):

PR #268, opened 2023-02-20
Enable Rubocop in CI

bastelfreak:


beaker-hostgenerator.gemspec
```diff
@@ -19,6 +19,8 @@ eos
   s.executables   = `git ls-files -- bin/*`.split("\n").map{ |f| File.basename(f) }
   s.require_paths = ["lib"]
 
+  s.required_ruby_version = Gem::Requirement.new('>= 2.4')
```
bastelfreak, line 22: this is what our CI currently tests on, so I assume that's the oldest supported version.

smortex (approved):
I did not checked `.rubocop_todo.yml` assuming the PR wants to focus on the Rubocop integration and the fixes will be part of future work, and this looks good to me.

PR #269, opened 2023-02-20
Remove deprecated genconfig2 script

ekohl:
While the warning suggested it would be removed in 1.x, we're already at 1.12 so there should have been plenty of time.


smortex (approved):
LGTM :smile:

bastelfreak:
same as https://github.com/voxpupuli/beaker-hostgenerator/pull/236

ekohl:
I should have checked that before.

PR #270, opened 2023-02-20
Drop fixup_node implementation

ekohl:
This was only used for EL 4 and a specific platform version, which are EOL now.


smortex (approved):
LGTM

PR #271, opened 2023-02-20
Drop nfs_server and consoleport defaults

ekohl:
Beaker defaults to consoleport 443, so specifying this is redundant and extremely noisy.

As for nfs_server, I could not find any code that uses it. It shows up in many markdown files and configs, but I've yet to find actual Ruby code that uses it.

### DIFF
--- a/lib/beaker-hostgenerator/data.rb
+++ b/lib/beaker-hostgenerator/data.rb
@@ -56,10 +56,7 @@ module BeakerHostGenerator
 
     BASE_CONFIG = {
       'HOSTS' => {},
-      'CONFIG' => {
-        'nfs_server' => 'none',
-        'consoleport' => 443,
-      }
+      'CONFIG' => {},
     }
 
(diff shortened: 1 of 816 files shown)


bastelfreak:
mhm maybe that's only used internally at Puppet?

ekohl:
Possible, but I think the should fix the tool to have a default then instead of requiring it to be specified everywhere.

smortex (approved):
The beginning looks okay, but my browser crash when I try to scroll down :-D  Adding `.diff` at the end of the URI helps.

This look reasonable to me.

ekohl:
I should have mentioned that it was mostly:
```
find test/fixtures -type f -exec sed -i '/consoleport: 443/d' {} +
find test/fixtures -type f -exec sed -i '/nfs_server: none/d' {} +
```

Then I fixed up 3 fixtures where `CONFIG` became `nil` instead of `{}`.

PR #273, opened 2023-02-21
Drop Ruby 2.4/2.5/2.6 support

bastelfreak:
All three versions are EoL, supporting them makes Ruby 3.2 support tricky and beaker itself now requires Ruby 2.7 or newer.


sebastianrakel (approved):

.rubocop_todo.yml
```diff
@@ -366,17 +375,18 @@ Style/ConditionalAssignment:
...
-# Cop supports --auto-correct.
+# This cop supports safe autocorrection (--autocorrect).
 Style/DefWithParentheses:
   Exclude:
     - 'lib/beaker-hostgenerator/hypervisor.rb'
 
-# Offense count: 10
+# Offense count: 11
 # Configuration parameters: AllowedConstants.
 Style/Documentation:
```
ekohl, line 385: Isn't this effectively all our files?

bastelfreak:
```
$ find lib/ | wc -l
20
```
at least half of it

PR #281, opened 2023-02-23
Fix multiple rubocop violations

bastelfreak:
I created one commit per cop. All of them are from the group of 'safe autocorrectable cops', not from the 'unsafe autocorrectable group' and not from the 'not autocorrectable group'.


bastelfreak:
I would like to merge https://github.com/voxpupuli/beaker-hostgenerator/pull/275 first

PR #287, opened 2023-03-08
Install hostname on CentOS 9 Stream in docker

evgeni:
This is not available in the minimal CentOS 9 Stream containers and specinfra relies on this for the hostname -f command to determine the FQDN.

Fixes: b7fe509f83284e85875108a48603abdd13ff3434


bastelfreak (approved):
I think as a workaround this is fine. on a longterm specinfra should probably switch to hostnamectl .

PR #303, opened 2023-05-02
use centos stream 8 box from cloud.centos.org

evgeni:
the one on vagrant cloud is old and outdated


ekohl (approved):
I considered doing this when I did the same for C9 but ended up not doing it. Probably too busy with getting C9 to just work in the first place.

PR #307, opened 2023-05-05
rubocop: Fix Lint violations

bastelfreak:


bastelfreak:
cherry-picked the changes from https://github.com/voxpupuli/beaker-hostgenerator/pull/291

lib/beaker-hostgenerator/hypervisor.rb
```diff
@@ -84,7 +84,7 @@ def global_config()
       #
       # @param [Integer] bhg_version The version of OS info to use when building
       #                              up the node definition.
-      def generate_node(node_info, base_config, bhg_version)
```
rwaffen, line 87: why define parameters when you plan to never use them? (_ syntax)

rwaffen (approved):

lib/beaker-hostgenerator/hypervisor.rb
bastelfreak: there are different implementations of this class, for example for hcloud, vagrant, docker. They all have to provide the method `generate_node` with the same signature. This is just the dummy/base implementation. I think it has to have the same signature as well.

PR #312, opened 2023-07-14
Fix missing include for `strscan`

jonathannewman:
StringScanner is used without a corresponding require.  The lack of this require breaks under Ruby 3.1 in some environments.

### DIFF
--- a/lib/beaker-hostgenerator/parser.rb
+++ b/lib/beaker-hostgenerator/parser.rb
@@ -1,6 +1,7 @@
 require 'beaker-hostgenerator/data'
 require 'beaker-hostgenerator/exceptions'
 require 'cgi'
+require 'strscan'
 
 module BeakerHostGenerator
   # Functions for parsing the raw user input host layout string and turning


smortex (commented):
LGTM.  Can you please rebase your changes on top of the master branch?

From your working directory:

```
git fetch origin         # Download the latest code we have here
git rebase origin/master # Move your commits on top of the main branch
git push -f              # Send the changes (-f is required because we re-wrote history)
```

CI should then pass :wink:

jonathannewman:
Rebased.

bastelfreak:
@jonathannewman can you tell us when/how this breaks? We havent noticed issues in the past.

ekohl (approved):
At least on Ruby 3.2 you can reproduce that it's not included by default:
```console
$ ruby -e 'StringScanner'  
-e:1:in `<main>': uninitialized constant StringScanner (NameError)
                                         
StringScanner                            
^^^^^^^^^^^^^
$ ruby -r strscan -e 'StringScanner'  
```

So I think this is correct and at worst doesn't hurt.

jonathannewman:
It is causing issues for me running beaker tests on my local system on the Mac with Ruby 3.1 and 3.2.  
```
/Users/jonathannewman/Documents/workspace/forked/me/pe-puppet-server-extensions/vendor/bundler/ruby/3.1.0/gems/beaker-hostgenerator-1.18.1/lib/beaker-hostgenerator/parser.rb:213:in `settings_string_to_map': uninitialized constant BeakerHostGenerator::Parser::StringScanner (NameError)

      stringscan = StringScanner.new(host_settings)
                   ^^^^^^^^^^^^^
Did you mean?  StringIO
```

It all depends on what was included before this code is run as to if it is a problem.  Adding in the require makes it more robust and obvious (and I'd argue more correct).

We have linux systems running the same tests that don't encounter this error, so it is system dependent.

ekohl:
> It all depends on what was included before this code is run as to if it is a problem. Adding in the require makes it more robust and obvious (and I'd argue more correct).
> 
> We have linux systems running the same tests that don't encounter this error, so it is system dependent.

This is the "joy" of the Ruby include model. I hate it with a passion and much prefer the explicit model of Python where every file is explicit in what it uses. Of course it doesn't work well with the Ruby namespace model, but I'd argue that the Ruby namespace model is also weaker than Python's.

PR #318, opened 2023-07-30
Debian: Install iproute2

bastelfreak:
Install iproute2. This provides `ss`:
https://packages.debian.org/bookworm/amd64/iproute2/filelist This is used by serverspec to check for open ports. This is probably not the best space because it's docker specific, but it will fix our tests.


smortex (approved):
LGTM

PR #320, opened 2023-08-02
Use centos/stream9 vagrant box

ekohl:
These are finally published on Vagrant Cloud which means it can automatically retrieve the latest version.

When I just tested this elsewhere I couldn't get the latest image to boot, so perhaps this isn't ready yet. Marking as draft until I can verify it works.


ekohl:
Current image is broken due to https://bugzilla.redhat.com/show_bug.cgi?id=2227722

PR #322, opened 2023-08-04
Drop legacy EL6 data

bastelfreak:
Result of a discussion we had in https://github.com/voxpupuli/beaker-hostgenerator/pull/321


ekohl:
Perhaps Puppet folks can weigh in here.

ekohl (changes requested):
I'd actually like to continue using `oracle`. Writing out `oraclelinux` is tedious. We didn't change `centos` to `centoslinux` and also don't write `centosstream`.

In puppet_metadata we can make sure to generate the right name.

bastelfreak:
I cleaned it up for the future. I think that's legacy code that could be removed with the next major release.

PR #327, opened 2023-09-08
(RE-15156) Add redhat-9-arm64 support

yachub:
I'm not sure about all the places that the `abs` and `vmpooler` platform data keys are used, but was a bit confused because the `redhat8-AARCH64` specifies the x86 template name `redhat-8-x86_64`, is that a mistake? https://github.com/voxpupuli/beaker-hostgenerator/blob/da8ac9a25030480925b79a84a34444d7f9d14970/lib/beaker-hostgenerator/data.rb#L348

So for the new platform's `vmpooler` key I used the expected name that we would use in vmpooler for that platform, is that correct? Or should that key just be omitted?

**EDIT**: Ah it looks like those keys related to the hypervisor gem. The above definitely seems like a bug per:
```
> beaker-hostgenerator redhat8-AARCH64 --hypervisor vmpooler
---
HOSTS:
  redhat8-AARCH64-1:
    platform: el-8-aarch64
    template: redhat-8-x86_64
    hypervisor: vmpooler
    roles:
    - agent
CONFIG:
  pooling_api: https://vmpooler-prod.k8s.infracore.puppet.net/
```

### DIFF
--- a/lib/beaker-hostgenerator/data.rb
+++ b/lib/beaker-hostgenerator/data.rb
@@ -361,6 +361,17 @@ module BeakerHostGenerator
                           'platform' => 'el-9-x86_64',
                         },
                       },
+                      'redhat9-AARCH64' => {
+                        general: {
+                          'platform' => 'el-9-aarch64',
+                        },
+                        abs: {
+                          'template' => 'redhat-9-arm64',
+                        },
+                        vmpooler: {
+                          'template' => 'redhat-9-arm64',
+                        },
+                      },
                       'sles11-32' => {
                         general: {
                           'platform' => 'sles-11-i386',
--- a/test/fixtures/generated/default/redhat9-AARCH64m
+++ b/test/fixtures/generated/default/redhat9-AARCH64m
@@ -1,0 +1,15 @@
+---
+arguments_string: redhat9-AARCH64m
+environment_variables: {}
+expected_hash:
+  HOSTS:
+    redhat9-AARCH64-1:
+      platform: el-9-aarch64
+      template: redhat-9-arm64
+      hypervisor: vmpooler
+      roles:
+      - agent
+      - master
+  CONFIG:
+    pooling_api: https://vmpooler-prod.k8s.infracore.puppet.net/
+expected_exception:
--- a/test/fixtures/generated/multiplatform/centos7-64a-redhat9-AARCH64-centos7-64aulcdfm
+++ b/test/fixtures/generated/multiplatform/centos7-64a-redhat9-AARCH64-centos7-64aulcdfm
@@ -1,0 +1,32 @@
+---
+arguments_string: centos7-64a-redhat9-AARCH64-centos7-64aulcdfm
+environment_variables: {}
+expected_hash:
+  HOSTS:
+    centos7-64-1:
+      platform: el-7-x86_64
+      hypervisor: vmpooler
+      template: centos-7-x86_64
+      roles:
+      - agent
+    redhat9-AARCH64-1:
+      platform: el-9-aarch64
+      template: redhat-9-arm64
+      hypervisor: vmpooler
+      roles:
+      - agent
+    centos7-64-2:
+      platform: el-7-x86_64
+      hypervisor: vmpooler
+      template: centos-7-x86_64
+      roles:
+      - agent
+      - ca
+      - classifier
+      - dashboard
+      - database
+      - frictionless
+      - master
+  CONFIG:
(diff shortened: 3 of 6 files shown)


ekohl:
For RHEL I haven't looked at correct generation yet. As you noticed, the data was inconsistent. Like `redhat7-AARCH64` uses `centos-7-arm64` in abs.

Ideally we would generate the ABS template in https://github.com/voxpupuli/beaker-hostgenerator/blob/master/lib/beaker-hostgenerator/hypervisor/abs.rb and vmpooler in https://github.com/voxpupuli/beaker-hostgenerator/blob/master/lib/beaker-hostgenerator/hypervisor/vmpooler.rb. Then it would be reduced to just listing the versions in [generate_os_info](https://github.com/voxpupuli/beaker-hostgenerator/blob/da8ac9a25030480925b79a84a34444d7f9d14970/lib/beaker-hostgenerator/data.rb#L1047).

The biggest problem is that Vox Pupuli doesn't use ABS or vmpooler so we really don't know what's available and what makes sense.

joshcooper:
Seems like a bug to me too. IMO BHG should always return a template/platform name that is consistent with the key (BHG string):

We do cross compile on some platforms, but the decision about whether to cross-compile and which build host to use lives in the platform definition for the vanagon project, for example: https://github.com/puppetlabs/puppet-runtime/blob/6c8bbc664aea36cbadc12ab0d62773250faf70f1/configs/platforms/sles-12-ppc64le.rb#L31-L32

In other words, I don't see a use case for BHG ever returning an Intel-based `template` for an ARM-based key.

joshcooper (approved):

joshcooper:
Not sure why, but I'm having trouble getting this working. This line is causes an error if `platform` is not defined:

https://github.com/voxpupuli/beaker-hostgenerator/blob/013dab5a0ab39c5d1b7383ec1265c382eb325dac/lib/beaker-hostgenerator/hypervisor/abs.rb#L27C39-L27C50

```
$ bx beaker-hostgenerator redhat9-ARM64a
bundler: failed to load command: beaker-hostgenerator (/home/josh/.rbenv/versions/2.7.6/lib/ruby/gems/2.7.0/bin/beaker-hostgenerator)
Traceback (most recent call last):
...
	 6: from /home/josh/work/beaker-hostgenerator/bin/beaker-hostgenerator:10:in `<top (required)>'
	 5: from /home/josh/work/beaker-hostgenerator/lib/beaker-hostgenerator/cli.rb:174:in `execute!'
	 4: from /home/josh/work/beaker-hostgenerator/lib/beaker-hostgenerator/cli.rb:168:in `execute'
	 3: from /home/josh/work/beaker-hostgenerator/lib/beaker-hostgenerator/generator.rb:30:in `generate'
	 2: from /home/josh/work/beaker-hostgenerator/lib/beaker-hostgenerator/generator.rb:30:in `each'
	 1: from /home/josh/work/beaker-hostgenerator/lib/beaker-hostgenerator/generator.rb:52:in `block in generate'
/home/josh/work/beaker-hostgenerator/lib/beaker-hostgenerator/hypervisor/vmpooler.rb:22:in `generate_node': undefined method `gsub' for nil:NilClass (NoMethodError)
```

I get the same when passing `abs` or `vmpooler` as the `hypervisor`

yachub:
@joshcooper Following precedent of most recently added platforms I used "AARCH64" instead of "ARM64". Does that work for you?

```bash
> beaker --version
      wWWWw
      |o o|
      | O |  5.3.1!
      |(")|
     / \X/ \
    |   V   |
    |   |   |
    
> beaker-hostgenerator -v
2.4.0

> beaker-hostgenerator redhat9-AARCH64a --hypervisor abs
---
HOSTS:
  redhat9-AARCH64-1:
    platform: el-9-aarch64
    template: redhat-9-arm64
    hypervisor: abs
    roles:
    - agent
CONFIG: {}
```

ekohl:
Still, would be good to either silently ignore unknown architectures or fail with a clear error instead of a stack trace.

joshcooper:
> Still, would be good to either silently ignore unknown architectures or fail with a clear error instead of a stack trace.

yeah that could be handled better. Took me a bit to understand what was going on. I also run into errors like that when trying to run beaker tests and I've forgotted to do `bundle update` to pull in a version of BHG that recognizes the new platform.

> I used "AARCH64" instead of "ARM64". Does that work for you?

Yeah I think it's good for the architecture to be consistent for all versions of the same OS. As for why we use different names for one OS vs another, I can only refer to https://twitter.com/lorenc_dan/status/1699794514586763408:

F5biGZgbcAA9RZ9

PR #248, opened 2022-04-29
Add Ubuntu 22.04

jkroepke:
Note:
```
bundle exec rake generate:fixtures 
```
generates tons of new files and changes. I only include the relevant changes for this PR

PR #328, opened 2023-09-08
(RE-15206) Add ubuntu-2204-arm64 fixtures

yachub:
I'm not sure if this is needed since there was some data added in https://github.com/voxpupuli/beaker-hostgenerator/pull/248 and released in [1.14.0](https://github.com/voxpupuli/beaker-hostgenerator/blob/master/CHANGELOG.md#1140-2022-05-30), but without any changes to data.rb, these were the fixtures generated with `bundle exec rake generate:fixtures`

Side note, it seems a bit odd that we don't set the template key for the ubuntu platforms here. This must be already be inserted by [beaker-abs](https://github.com/puppetlabs/beaker-abs), but I haven't gone looking

### DIFF
--- a/test/fixtures/generated/default/ubuntu2204-AARCH64m
+++ b/test/fixtures/generated/default/ubuntu2204-AARCH64m
@@ -1,0 +1,14 @@
+---
+arguments_string: ubuntu2204-AARCH64m
+environment_variables: {}
+expected_hash:
+  HOSTS:
+    ubuntu2204-AARCH64-1:
+      platform: ubuntu-22.04-aarch64
+      hypervisor: vmpooler
+      roles:
+      - agent
+      - master
+  CONFIG:
+    pooling_api: https://vmpooler-prod.k8s.infracore.puppet.net/
+expected_exception:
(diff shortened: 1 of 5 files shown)


ekohl:
> I'm not sure if this is needed since there was some data added in #248 and released in [1.14.0](https://github.com/voxpupuli/beaker-hostgenerator/blob/master/CHANGELOG.md#1140-2022-05-30), but without any changes to data.rb, these were the fixtures generated with `bundle exec rake generate:fixtures`

It's a good question. As more logic is generated dynamically, testing it becomes more important. But just storing huge generated files doesn't guarantee it's correct.

> Side note, it seems a bit odd that we don't set the `template` key for the ubuntu platforms here. This must be already be inserted by [beaker-abs](https://github.com/puppetlabs/beaker-abs), but I haven't gone looking:

I've been working on moving away from a huge `data.rb` and instead push logic into specific hypervisors to generate data dynamically. So https://github.com/voxpupuli/beaker-hostgenerator/blob/master/lib/beaker-hostgenerator/hypervisor/abs.rb already does this.

mhashizume (approved):

PR #331, opened 2023-09-11
(RE-15541) Add debian-12-arm64 support

yachub:


ekohl (changes requested):
Same notes as in https://github.com/voxpupuli/beaker-hostgenerator/pull/330.

PR #345, opened 2023-11-17
(PA-5948) Update host generator to use aarch64 for amazon linux

span786:
Updated data.rb in host-generator to use 'aarch64' like it is using everywhere in project

### DIFF
--- a/lib/beaker-hostgenerator/data.rb
+++ b/lib/beaker-hostgenerator/data.rb
@@ -1111,7 +1111,7 @@ module BeakerHostGenerator
 
       # Amazon Linux
       yield %w[amazon2023-64 amazon-2023-x86_64]
-      yield %w[amazon2023-AARCH64 amazon-2023-arm64]
+      yield %w[amazon2023-AARCH64 amazon-2023-aarch64]
 
       # AlmaLinux and Rocky
       %w[almalinux rocky].each do |os|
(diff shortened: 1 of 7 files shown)


mhashizume (approved):

mhashizume (changes requested):
Sorry, I think we need one more change here. The goal is to have BHG generate a `platform` that is appropriate for our Vanagon projects, but a `template` that is appropriate for ABS/VMPooler. Right now, your PR generates this:

```
bundle exec beaker-hostgenerator amazon2023-AARCH64a --hypervisor abs
---
HOSTS:
  amazon2023-AARCH64-1:
    platform: amazon-2023-aarch64
    hypervisor: abs
    template: amazon-2023-aarch64
    roles:
    - agent
CONFIG: {}
```

For our CI, we need the `template` to match what's in ABS/VMPooler, which is `amazon-2023-arm64`.

In addition to what you already have here, we need to move Amazon to this regex line: https://github.com/voxpupuli/beaker-hostgenerator/blob/64f315292da95440a025223a47b57d767cdd7f0f/lib/beaker-hostgenerator/hypervisor/abs.rb#L30

So we get this desired result:

```
bundle exec beaker-hostgenerator amazon2023-AARCH64a --hypervisor abs
---
HOSTS:
  amazon2023-AARCH64-1:
    platform: amazon-2023-aarch64
    hypervisor: abs
    template: amazon-2023-arm64
    roles:
    - agent
CONFIG: {}
```

mhashizume (approved):

mhashizume:
LGTM, I'm not sure what we need to do to make Codecov happy

bastelfreak:
@mhashizume just ignore it :)

PR #350, opened 2024-01-23
Fix rhel9 power ABS name

nmburgan:
On behalf of @trvs-sdlr


trvs-sdlr (approved):

ekohl (approved):
> I assume the codecov thing is not such a big deal since the previous line didn't have coverage either.

I honestly don't care about code coverage at all in general, but in this particular case it can't measure the coverage by integration tests either so it's really misleading. Like in this case, it's often misleading. With 100% coverage you can easily have bugs in the integration.

PR #356, opened 2024-02-13
Drop code coverage

joshcooper:


ekohl (approved):

ekohl:
For context: it was broken. It doesn't take the tests into account. Probably because this gem has both rspec and minitest configured.

So we either need to migrate all our minitest tests to rspec or drop them. There's certainly something to be said for unifying on a single test suite, but I never really cared about code coverage.

PR #359, opened 2024-03-13
Add redhat support to docker hypervisor

h0tw1r3:
Vendor provides [universal base images](https://catalog.redhat.com/software/base-images) with a permissive [EULA](https://developers.redhat.com/articles/ubi-faq#legal_and_licensing).

> Does UBI let me distribute my container images anywhere I want?
> 
> Yes. Building on UBI means Red Hat is never going to tell you where or how you can distribute your images. You can distribute your images wherever and however you like.

### DIFF
--- a/lib/beaker-hostgenerator/hypervisor/docker.rb
+++ b/lib/beaker-hostgenerator/hypervisor/docker.rb
@@ -17,7 +17,7 @@ module BeakerHostGenerator
         end
 
         base_config['image'] = image(ostype)
-        base_config['image'].prepend('amd64/') if node_info['bits'] == '64' && !base_config['image'].start_with?('quay.io/')
+        base_config['image'].prepend('amd64/') if node_info['bits'] == '64' && !base_config['image'].start_with?('quay.io/', 'registry.access.redhat.com/')
 
         base_generate_node(node_info, base_config, bhg_version, :docker)
       end
@@ -32,6 +32,9 @@ module BeakerHostGenerator
           version = ostype.delete_prefix('centos')
           tag = (version.to_i >= 8) ? "stream#{version}" : "centos#{version}"
           image = "quay.io/centos/centos:#{tag}"
+        when /^redhat/
+          version = ostype.delete_prefix('redhat')
+          image = "registry.access.redhat.com/ubi#{version}/ubi-init:latest"
         when /^oracle/
           image.sub!(/\w+/, 'oraclelinux')
         when /^opensuse/
@@ -49,7 +52,7 @@ module BeakerHostGenerator
 
       def image_commands(ostype)
         case ostype
-        when /^(almalinux|centos|rocky|oracle)/
+        when /^(almalinux|centos|rocky|oracle|redhat)/
           [
             'cp /bin/true /sbin/agetty',
             el_package_install_command(ostype.delete_prefix(Regexp.last_match(1)).to_i),


lib/beaker-hostgenerator/hypervisor/docker.rb
```diff
@@ -17,7 +17,7 @@ def generate_node(node_info, base_config, bhg_version)
         end
 
         base_config['image'] = image(ostype)
-        base_config['image'].prepend('amd64/') if node_info['bits'] == '64' && !base_config['image'].start_with?('quay.io/')
+        base_config['image'].prepend('amd64/') if node_info['bits'] == '64' && base_config['image'] !~ /^(quay\.io|registry\.access\.redhat\.com)/
```
ekohl, line 20: You can give multiple prefixes to `start_with?` so I'd expect this to work:
```suggestion
        base_config['image'].prepend('amd64/') if node_info['bits'] == '64' && !base_config['image'].start_with('quay.io', 'registry.access.redhat.com')
```

ekohl (commented):
Nice. Do you have a PR somewhere that tests this out?

bastelfreak:
In the past we enabled acceptance testing a few times when the OSes were already in the metadata.json. That's a bit tricky because that can break CI pipelines because there are unexpected new jobs for some OSes. We've RedHat listed quite often. I would love to see this merged, but we need to more sure that it passes on our modules and that the registry can handle some concurrent pulls (quay.io seems to be rather flaky :( )

h0tw1r3:
Will this impact existing pipelines until the puppet_metadata change is merged?

bastelfreak:
this will impact them as soon as we release beaker-hostgenerator and puppet_metadata as minor versions.

ekohl:
Looks like it fails because chrony/ntpdate isn't present in UBI. I knew it had a slightly filtered list and I guess this is one of them.

This means we need to update the packages to install.

Right now this is determined in a complex way:

* `host_packages(host)` determines which packages to install: https://github.com/voxpupuli/beaker/blob/0b477ea3f84bfdd053bad614f7a874ffbd233cf1/lib/beaker/host_prebuilt_steps.rb#L106-L153 
* This is used in beaker-docker: https://github.com/voxpupuli/beaker-docker/blob/72dda38788e291394271bc186d7e20eb6ee1d241/lib/beaker/hypervisor/docker.rb#L498

I'm not sure what the best way to filter it out is without breaking non-Docker deployments. Perhaps reject some values in beaker-docker for RHEL?

Then we also find a way to exclude it from the testing matrix in https://github.com/voxpupuli/puppet-chrony. Perhaps some options to add in puppet_metadata?

h0tw1r3:
> Looks like it fails because chrony/ntpdate isn't present in UBI. I knew it had a slightly filtered list and I guess this is one of them.
> 
> This means we need to update the packages to install.
> 
> Right now this is determined in a complex way:
> 
>     * `host_packages(host)` determines which packages to install: https://github.com/voxpupuli/beaker/blob/0b477ea3f84bfdd053bad614f7a874ffbd233cf1/lib/beaker/host_prebuilt_steps.rb#L106-L153
> 
>     * This is used in beaker-docker: https://github.com/voxpupuli/beaker-docker/blob/72dda38788e291394271bc186d7e20eb6ee1d241/lib/beaker/hypervisor/docker.rb#L498
> 

I'm testing locally with a modification to https://github.com/voxpupuli/beaker/blob/0b477ea3f84bfdd053bad614f7a874ffbd233cf1/lib/beaker/host_prebuilt_steps.rb#L110 which filters time related packages if the host timesync option is disabled (which is a default I think).

> I'm not sure what the best way to filter it out is without breaking non-Docker deployments. Perhaps reject some values in beaker-docker for RHEL?
> 
> Then we also find a way to exclude it from the testing matrix in https://github.com/voxpupuli/puppet-chrony. Perhaps some options to add in puppet_metadata?

Yes, filtering will need to be added. I suggest it be left up to the specific modules workflow as I suspect it will be a moving target over time.

h0tw1r3:
> I would like to see all the packages in beaker-docker moved to beaker. They really aren't docker specific.

Agreed, it shouldn't be needed there as it's done in host_prebuilt_steps for all hypervisors.

ekohl:
> I would like to see all the packages in beaker-docker moved to beaker. They really aren't docker specific.

As I described: we already have that since https://github.com/voxpupuli/beaker-docker/commit/e126b7005475bd4e4a8508be4a43a1e9287b2163.

IMHO there is a benefit to installing packaging during container building: they're cached so if you rerun tests you don't need to install them again. It would be even better if we publish the built containers for even faster CI.

Though you did trigger me to look at `Beaker::HostPrebuiltSteps` and in `validate_host` it ensures the host packages are present. So just changing the definition in beaker-docker is probably insufficient.

That makes me think: can we actually refactor it so it's only installed if you actually timesync and do that for every OS? I opened https://github.com/voxpupuli/beaker/pull/1853 as a start for that.

kenyon:
Replaced by #459.

PR #353, opened 2024-01-27
Use latest centos boxes instead of specific dates

ekohl:
The centos/stream8 and centos/stream9 images are still not correct on Vagrant cloud, so it needs to be provided via a box_url. The centos/stream9 box URL is now a 404 so this changes it to latest.

PR #374, opened 2024-06-18
Use latest CentOS boxes instead of specific dates

daberkow:
I added tests to #353 to see if that does what we need, not sure proper etiquette here, if we want me to PR into the other branch then merge that one?


daberkow:
@bastelfreak, this is another one that will help get selinux back on the main branches and happy!

bastelfreak (approved):

PR #412, opened 2025-05-30
Fix template for windows 2025

skyamgarp:


skyamgarp:
> `ent` was removed in https://github.com/voxpupuli/beaker-hostgenerator/pull/410/files, now you are adding it back. How can we verify what's actually correct?

https://github.com/voxpupuli/beaker-hostgenerator/pull/410#discussion_r2052636481
It's breaking because the template is not fixed. Hence adding it again.

joshcooper:
We don't want to add it back. We need to update the jobs where bhg is called with the old name.

PR #430, opened 2025-08-19
Add OracleLinux 10 support

bastelfreak:


bastelfreak:
tested in https://github.com/voxpupuli/puppet-keepalived/pull/363

PR #432, opened 2025-09-01
Generate Rocky, Alma and RedHat dynamically

traylenator:
Loop over major numbers for Rocky, Alma and RedHat


traylenator:
> Thanks! Nice improvement, even if the vmpooler bits are weird. But those were already weird before.

For sure there was a lot of stuff that made no sense to me. Might strip  EL7 out which would remove quite a few weird cases.
SLES look easy.
Windows needs a bit more for the locale parameters.

PR #433, opened 2025-09-06
Use a loop for SLES

traylenator:
Refactor SLES host definitions for easier management.


traylenator:
Don't let my data preservation colleagues hear that. 😄

Steve
________________________________

PR #461, opened 2026-04-13
add oracle10 vagrant box

marcusdots:


kenyon (approved):

marcusdots:
The logic does not work. The images from bento/ only support virtualbox (`BEAKER_HYPERVISOR=vagrant`), but the github pipeline runs on libvirt (`BEAKER_HYPERVISOR=vagrant_libvirt`). We need to find new images.

lib/beaker-hostgenerator/hypervisor/vagrant.rb
```diff
@@ -16,12 +16,40 @@ class Vagrant < BeakerHostGenerator::Hypervisor::Interface
...
+      ROCKY_LINUX_BOXES = {
+        # there is no generic/rocky10, so we introduce this workaround
+        # bento/ does not support libvirt
+        '10' => 'cloud-image/rocky-10',
+        '9' => 'generic/rocky9',
+        '8' => 'generic/rocky8',
+      }.freeze
+
+      ORACLE_LINUX_BOXES = {
+        '10' => 'cloud-image/rocky-10', # We know this is wrong, but there is no Oracle-10
```
ekohl, line 28: Then IMHO it should not yield a result and raise an error.

ekohl (commented):
Is there a reason not to switch all images to `cloud-image`, regardless of version?

lib/beaker-hostgenerator/hypervisor/vagrant.rb
marcusdots: We may try https://portal.cloud.hashicorp.com/vagrant/discover/crystax/oracle10

marcusdots: > Is there a reason not to switch all images to cloud-image, regardless of version?

I did not want to break existing functionality by switching images, but switching to cloud-image might be an easy way forward.

